PR #1, opened 2016-12-20
Optimise the icons

vinkla:
I've optimised the icons with ImageAlpha and ImageOptim and saved around 40%.


vinkla:
The look as before but with less weight.

JeffreyWay:
Thank ya!

PR #4, opened 2016-12-23
Implement cross-env, fixes #3

### LINKED ISSUE #3: Use cross-env for NODE_ENV
Haven't looked through everything yet, but in your scripts your are using

```json
"scripts": {
  "webpack": "webpack --progress --hide-modules",
  "dev": "webpack --watch --progress --hide-modules",
  "hmr": "webpack-dev-server --inline --hot",
  "production": "export NODE_ENV=production && webpack --progress --hide-modules"
}
```

Where export is not supported in non-UNIX platforms. I suggest pulling in cross-env as additional dev dependency.

This way, I am able to run `yarn run production` and `npm run production` on windows.

PascaleBeier:
Works fine for me! Thanks for your hard work.


JeffreyWay:
Thanks!

PR #21, opened 2016-12-29
Add multiple css preprocessor support

brunjick:
Basically, I have vendor css framework, which is less, and my own styles, which is stylus. I needed a way to define both entries to have `vendor.css` and `app.css` separately. So I can now mix like this:
```js
mix
   .less(paths.assets + '/less/app.less', paths.public + '/css/vendor.css')
   // I have stylus declared in webpack.config.js like other preprocessors,
   // this PR doesn't include it
   .preprocess('stylus', paths.assets + '/stylus/app.styl', paths.public + '/css/app.css')
   // other stuff
```

PS. sorry for extra whitespace removal, editor did it automatically 😜

### DIFF
--- a/setup/webpack.config.js
+++ b/setup/webpack.config.js
@@ -86,7 +86,7 @@ module.exports.module = {
                 loaders: {
                     js: 'babel-loader' + Mix.babelConfig()
                 },
-                  
+
                 postcss: [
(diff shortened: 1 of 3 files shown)


brunjick:
Ok, this doesn't work well because css gets duplicated in both files

PR #33, opened 2017-01-04
Fix default public and cache paths when not using Laravel

kaidesu:
Fixes the default paths when not using Mix with a Laravel project.

Ran into this issue specifically with the `Mix.json` cache file using the path `.//Mix.json` by default.


jerguslejko:
When using `mix` in Laravel the trailing slash is not appended. Therefore, for consistent api, it should not be appended automatically when working outside Laravel.

PR #35, opened 2017-01-04
Fixed typo and added code quote

devanandb:
On line 43, fixed the typo (unless you really meant nom to run her ;) and hot reload :P :P).


JeffreyWay:
hahaha. Dang auto-correct.

PR #44, opened 2017-01-05
Fix: TypeError: Mix.root is not a function

kamalkhan:
`mix.copy` throws a **TypeError: Mix.root is not a function**

Fixed by adding a root method to *src/Mix.js*


kamalkhan:
Closing. Already fixed by #41

PR #72, opened 2017-01-10
Load files to/from"img" directory

Omranic:


jerguslejko:
What if the directory is named something else than `img`?

Omranic:
@jerguslejko same argument goes for "**js**" directory 🙂 isn't it better to save images in a dedicated directory rather than leaving them in the global namespace?! It's good to have conventions anyway. If you've better solution feel free to commit upon this PR or send another one 😉

vinkla:
I agree with @jerguslejko we don't want to force the naming of the directory.

Omranic:
Am I missing something? Isn't that the same as **js**, and **css** directories? Why should we leave images in the root `public` directory? Correct me if I'm wrong..

vinkla:
@Omranic But aren't those easy configurable through the `webpack.mix.js` file? Personally I prefer `styles`, `scripts` and `images` since it is more generic.

JeffreyWay:
Hey, everyone - I'm going to use a "/images" output folder name. If folks *really* need that to configurable, we can add an override or something.

nmfzone:
Huh, finally. #75

PR #73, opened 2017-01-11
Add support for Node.js 4.x

incon:


JeffreyWay:
I'm not sure that I see a huge benefit for Node 4 support. If tons of people request it, I can merge this. Otherwise, I'd prefer not to.

PR #97, opened 2017-01-14
Fixed failing tests

bvangraafeiland:
On Windows, the directory separator in the output became a double backslash, resulting in incorrect paths in the manifest file (when using the mix() function from Laravel). Furthermore, some test cases were outdated and did not pass anymore. I'm not fully sure about the chunkhash one though.


bvangraafeiland:
Not to sound impatient, but it seems like this change is undesired, could you elaborate? I'd be happy to fix what's missing, as in its current form, HMR simply doesn't work on Windows.

JeffreyWay:
Not ignoring you. Will go through it soon. :)

Eldair:
When is this going to be merged, L5.4 mix function is unusable when output paths in manifest are wrong.

PR #111, opened 2017-01-17
Few fixes and enhancements 

eblin:
Hey Jeffrey,

First of all great work on laravel-mix and everything you do in laracast :)

I'm transitioning our platform to laravel in 5.3 and decided to use laravel-mix for our project rather than elixir since I like webpack much better.

This PR fixes a few issues I encountered + also adds test (with coverage 100%)

I tried to be as descriptive as possible in the commits, but if you need more information about changes let me know.

Summary:
Testing – added tests for everything and code coverage of 100% now :)
Bug Fixes: 
- Be able to do multiple `mix.sass()` or `mix.less()`
- Allow `laravel-mix` to run as a locally installed module with `npm link`
- Object.values() does not work in latest LTS node (6.9.4) without the --harmony flag, this would break versioning.

I have a few more changes for docs that i need to write down specially a how to use laravel-mix in 5.3 which is not hard but not too obvious.

Thanks for all your work and I hope this PR helps.


src/Paths.js
```diff
@@ -1,20 +1,35 @@
...
+    mix() {
         return this.root('webpack.mix');
-    },
-
+    }
 
     /**
      * Determine the project root.
      *
+     * @param {string|null} rootPath
```
anchiceratops: rootPath isn't a `root()` param.

JeffreyWay:
Wow - a lot to work through here. Thanks so much!

About the multiple .sass() commit, yeah, the issue there is that it'll create the JavaScript files for each stylesheet as well, which we don't need. So I'll need to figure out how to remove those from the bundle.

Anyways, I'm going to start working through this.

src/Paths.js
eblin: Haha yes sorry I forgot to remove it when I refactored this to a class. I corrected it and also added a missing one in `Mix.detectHotReload`

eblin:
@JeffreyWay No problem! and yes I noticed the extra .js file as well, I will investigate further today and see if I can fix that.

JeffreyWay:
@eblin - Don't worry about it right now. I'm right in the middle of fixing it.

src/Paths.js
anchiceratops: @eblin awesome!

JeffreyWay:
Okay I think I've fixed it. 

I've merged in your commits. Thanks again!

eblin:
Hey @JeffreyWay I merged all your changes and everything works beautifully! I see you replaced ManifestPlugin with StatsWriter 👍 

Only thing that crashed was Object.values(), Object.values() doesn't work without --harmony flag, so the Manifest.js crashes on node v6.9.4 (latest LTS), this is not an issue in node v7+

https://github.com/JeffreyWay/laravel-mix/blob/master/src/Manifest.js#L20

Same thing was happening in `Versioning.js` the fix is really simple so I don't know if it merits another PR, here it is: https://github.com/JeffreyWay/laravel-mix/blob/master/src/Versioning.js#L3

JeffreyWay:
Argh. Forgot to swap that out, haha. Thanks will do.

PR #113, opened 2017-01-18
Add default npm scripts

cretueusebiu:
This PR allows to just use :

```json
"scripts": {
  "dev": "laravel-mix dev",  
  "watch": "laravel-mix watch",
  "hot": "laravel-mix hot",
  "production": "laravel-mix production"
},
```

in your package.json instead of the long commands.

### DIFF
--- a/bin/laravel-mix.js
+++ b/bin/laravel-mix.js
@@ -1,0 +1,34 @@
+#!/usr/bin/env node
+
+const script = process.argv[2];
+const spawn = require('cross-spawn');
+
+let args = ''
+
+switch (script) {
+    case 'dev':
+        args = ['NODE_ENV=development', 'webpack', '--progress', '--hide-modules'];
+        break;
+
+    case 'watch':
+        args = ['NODE_ENV=development', 'webpack', '--watch', '--progress', '--hide-modules'];
+        break;
+
+    case 'hot':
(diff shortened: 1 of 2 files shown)


bin/laravel-mix.js
```diff
@@ -0,0 +1,34 @@
...
+    case 'production':
+        args = ['NODE_ENV=production', 'webpack', '--progress', '--hide-modules'];
+        break;
+
+    default:
+        console.log(`Unknown script "${script}".`);
+        return;
+}
+
+args = args.concat(['--config=node_modules/laravel-mix/setup/webpack.config.js']);
```
DarkaOnLine: With this we would not be able to have custom config :(

cretueusebiu: In that case you would replace "dev": "laravel-mix dev"  with the long version command. But I think it could also be configurable this way too.

DarkaOnLine: So what're the benefits of having this short command? In the command line wee will execute `npm run dev` in both ways?

In the other hand, this will help to add or remove arguments inside the package and users will not have to update their `package.json` file, which is great.

JeffreyWay:
Thanks for this, but I think we'll hold off for now. I think folks will want to modify these scripts, and if we tuck them behind a `laravel-mix` executable, they won't know how.

PR #118, opened 2017-01-18
Fix notification on Windows

### LINKED ISSUE #114: Toast notification
With Laravel Elixir I got notifications after each task eg. "SASS compiled" and there was always an error message when something went wrong with the compilation  process.

Now though with Laravel Mix I always get the same meaningless toast notification, even if an error occurred.

notification-mix

"Nowe powiadomienie" translates to "New notification" so that's probably the default message provided by operating system. Is there a way to customize it?

If any of that matters, I'm using laravel-mix@0.4.0, Windows 10 and PowerShell as a built-in console in PhpStorm.

cretueusebiu:
Fixes #114

### DIFF
--- a/setup/webpack.config.js
+++ b/setup/webpack.config.js
@@ -268,7 +268,7 @@ if (Mix.notifications) {
         new plugins.WebpackNotifierPlugin({
             title: 'Laravel Mix',
             alwaysNotify: true,
-            contentImage: 'node_modules/laravel-mix/icons/laravel.png'
+            contentImage: path.join(__dirname, 'node_modules/laravel-mix/icons/laravel.png')
         })
     );
 }


KKSzymanowski:
@cretueusebiu Thanks a lot. This is probably not released yet, because I get the old version, but that's no problem.

However this works only when I copy the `webpack.config.js` to my root directory, when using the defaults in `node_modules` it still shows the previous message. I'm no expert in node but maybe we should do something like this:
```js
contentImage: __dirname.contains('node_modules') ?
        path.join(__dirname, '../icons/laravel.png') :
        path.join(__dirname, 'node_modules/laravel-mix/icons/laravel.png')
```

KKSzymanowski:
Sorry, `contains` doesn't work, `indexOf` however do:
```js
contentImage:  __dirname.indexOf('node_modules') >= 0 ?
    path.join(__dirname, '../icons/laravel.png') :
    path.join(__dirname, 'node_modules/laravel-mix/icons/laravel.png')
```

cretueusebiu:
Event better we can use `Mix.Paths.root('node_modules/laravel-mix/icons/laravel.png')`.

PR #121, opened 2017-01-19
[Bugfix] Return the final webpack config in `finalize` method and export it.

mathieutu:
Previously, in the `Mix.finalize` method, the merged webpack config was stored in `this.webpackConfig` in case of we provide a custom config. 
But we don't do anything with that `this.webpackConfig`, so the final webpack config (**`module.exports`** in `webpack.config.js`) **was not updated**.

Now, we return in any case the final webpack config (merged, or not) and export it to webpack.

Without that, I didn't succeed to use the `mix.webpackConfig({})` helper :
https://github.com/JeffreyWay/laravel-mix/blob/master/docs/quick-webpack-configuration.md

I may miss something, let me know in case.

Thanks for your work, hope mine will be useful.

### DIFF
--- a/setup/webpack.config.js
+++ b/setup/webpack.config.js
@@ -340,4 +340,4 @@ if (Mix.inProduction) {
  | If mix.webpackConfig() is called, we'll merge it in, and build!
  |
  */
-Mix.finalize(module.exports);
+module.exports = Mix.finalize(module.exports);
--- a/src/Mix.js
+++ b/src/Mix.js
@@ -49,7 +49,7 @@ class Mix {
      * @param {object} webpackConfig
      */
     finalize(webpackConfig) {
-        if (! this.webpackConfig) return;
+        if (! this.webpackConfig) return webpackConfig;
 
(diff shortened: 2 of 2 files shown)


JeffreyWay:
Hmm - I just tried it out, and mix.webpackConfig() did update `module.exports` as expected.

mathieutu:
Really? Weird... I've made a little example to show you with laravel-mix 0.5.0 : 

screen shot 2017-01-20 at 10 32 16

But in anyway I can't understand how it's supposed to work if it's working because I didn't found a piece of code to export the new config...

justkidding96:
@mathieutu Thanks for this solution. With this I answered my question asked here: https://github.com/JeffreyWay/laravel-mix/issues/126 I hope this is merged quickly inside the master @JeffreyWay ?

JeffreyWay:
Okay, I see what the problem was. A previous PR broke the merging. I've fixed it, and updated the test so that this won't happen again.

Thanks!

Issue #96, opened 2017-01-14
jQuery via ProvidePlugin issue


natocTo:
Hello, after this commit https://github.com/JeffreyWay/laravel-mix/commit/6e1ec3dcd4c3f4408c4ee119d70a1eb7ce6d450e I get error when compiling my assets.

The problem is that I do not use jquery but for example in Dropzone.js is this code.

``` javascript
if (typeof jQuery !== "undefined" && jQuery !== null) {
    jQuery.fn.dropzone = function(options) {
      return this.each(function() {
        return new Dropzone(this, options);
      });
    };
  }
```
so webpack require jquery but this ended with error.

Error:
```
Can't resolve 'jquery' in 'C:\xampp\htdocs\Moje\WEBSITES\VueShop\node_modules\dropzone\dist'
```

What about some provideJquery() function added to Laravel Mix for these who need this, or maybe there is other better solution.

Maybe is problem with Dropzone.js, how they determine if there is Jquery.

eblin:
The problem here is that most scripts will expect jQuery to be in the global `window` scope, the fix is to add `"window.jQuery": 'jquery'` to `ProvidePlugin` in `webpack.config.js`

```js
...
new webpack.ProvidePlugin({
        jQuery: 'jquery',
        $: 'jquery',
        jquery: 'jquery',
        "window.jQuery": 'jquery'
    }),
....
```
I had the same issue, I'll submit a PR soon, but if it doesn't get accepted this worked for me.

natocTo:
I have not jquery in my project. But because that check, webpack trying require that which end with error.

Maybe the best solution is copy webpack config and just remove this plugin when I do not need it.

JeffreyWay:
Yeah, hmm... will have to think about what to do here.

eblin:
Ah my bad sorry! I can't believe I missed that lol. 

Anyways, this would have to be configurable I guess, that way the entire object passed to ProvidePlugin could be set by the developer.

something like ... `Mix.autoloadModules({...})`

QWp6t:
I suggest using the following API:
```js
mix.autoload({  // or Mix.autoload() ?
  'jquery': ['$', 'window.jQuery', 'jQuery'],
  'tether': ['window.Tether', 'Tether'],
});
```
Where the key is the node module and the value is an array of vars to reference it.

thecrypticace:
I encountered this issue when using Laravel Echo too. I quite like the `mix.autoload` idea.

[...]

PR #124, opened 2017-01-20
Allow configuration of ProvidePlugin through “autoloading”

thecrypticace:
I've added `mix.autoload()` to configure webpack's ProvidePlugin. This should solve the problem in issue https://github.com/JeffreyWay/laravel-mix/issues/96 with libraries that use jQuery. h/t to @QWp6t in that thread for the example API which I've implemented here.

Example usage:
```js
mix.autoload({
  'jquery': ['$', 'window.jQuery', 'jQuery'],
});
```

I've got it on my todo list to add tests to verify this works. Just wanted to get a PR out there to start a conversation on the suggested API.

### DIFF
--- a/setup/webpack.config.js
+++ b/setup/webpack.config.js
@@ -234,11 +234,7 @@ module.exports.devServer = {
  */
 
 module.exports.plugins = (module.exports.plugins || []).concat([
-    new webpack.ProvidePlugin({
-        jQuery: 'jquery',
-        $: 'jquery',
-        jquery: 'jquery'
-    }),
+    new webpack.ProvidePlugin(Mix.js.autoload || {}),
 
     new plugins.FriendlyErrorsWebpackPlugin(),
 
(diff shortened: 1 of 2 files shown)


JeffreyWay:
We can do this...but I really hate that folks would be expected to call this method..

thecrypticace:
I do agree on that. I spent a little time trying to figure out a way to override ProvidePlugin but failed to find something helpful. Perhaps jQuery could be in the default set of autoloads and calling it with an empty `{}` could remove them? I'm still not sold on that but it's the best I can come up with at the moment.

eblin:
@thecrypticace @JeffreyWay  I think a good solution would be, have sensible defaults that work for the 80% of use cases. In this case have jquery be loaded by default like you said and pass the empty obj to remove it if you want. 

It would be very similar to how `Mix.babelConfig()` method works.

thecrypticace:
Updated.

bvangraafeiland:
Would be nice to have this, the way it is now you have to add jquery to package.json even if you don't use it.

PR #128, opened 2017-01-22
simplified the exports for a cleaner import in webpack.mix.js

bvangraafeiland:
Although this may have been a conscious design choice, it was a bit confusing for me personally to have the entire exports into a lowercase `mix` module, which also caused the default `webpack.mix.js` to have an unnecessary `.mix` suffixed to it. Again, if that was the intention, feel free to reject but for me it's just a bit clearer.


bvangraafeiland:
Sorry for being off topic, but in my previous PR I used `string.replace`, which turned out to replace only the first occurrence of the needle. The fixes are included in the other commit here.

JeffreyWay:
Thanks!

Issue #101, opened 2017-01-14
mix.copy does not allow for the glob params.


Landish:
laravel.mix.js

```js
mix.copy('src/images/*{.png,.jpg}', 'dist/images');
```
Let's say I have 2 files:  

```
/src/images/example.png
/src/images/example.jpg
```
when I run the `npm run webpack` it copies whole `images` dir into `dist`:

```
/dist/images/src/images/example.png
/dist/images/src/images/example.jpg
```

and should be:
```
/dist/images/example.png
/dist/images/example.jpg
```

JeffreyWay:
Hmm - maybe update to the latest version of Mix and try again. I just tested this out in a new project, and it worked. I did...

```js
mix.copy('tests/Unit/*{.txt,.php}', 'somewhere/else');
```

...and it only copied txt and php files.

Landish:
@JeffreyWay Nope! 

Still having that issue on version 0.5.1.

I even tried to implement directory structure according to your example and ended up with this:

<img width="238" alt="screen shot 2017-01-22 at 20 48 26" src="https://cloud.githubusercontent.com/assets/534610/22184065/5cde14e4-e0e4-11e6-9b5b-60e329bb03d9.png">

zgabievi:
+1 same issue to me.

![screen shot 2017-01-23 at 23 40 55](https://cloud.githubusercontent.com/assets/1515299/22219812/baca9b90-e1c5-11e6-97ca-d5ab6905e6d4.jpg)

JeffreyWay:
This seems to be an issue with the copy plugin we're using. 

https://github.com/kevlened/copy-webpack-plugin/issues/73

Looks like it may be fixed in the next release.

JeffreyWay:
This is fixed now in #135.

ianhk:
I'm not sure about this fix - it means you can't copy a tree without it being flattened. 

My solution is to use the 'context' copy-webpack-plugin option. Context allows you to specify the root that the copy is relative to - rather than implying it all from the 'from' arg.

Something like: 

`mix.copy2 = function (from, to, root) {
    var options = {
        from: from,
        to: Mix.Paths.root(to)
    };
    if (root) {
        options.context = root;
    }
    Mix.copy = (Mix.copy || []).concat(options);
};`

njt1982:
Just had a very similar problem myself:

    .copy('resources/assets/app/**/*.html', 'public/templates')

This ends up making a path like this:

[...]

PR #135, opened 2017-01-23
copy command should have flatten paths. fix for issue #101

zgabievi:
As we read in documentation of [copy-webpack-plugin](https://github.com/kevlened/copy-webpack-plugin), property **flatten**:

> Removes all directory references and only copies file names
> If files have the same name, the result is non-deterministic


JeffreyWay:
Excellent! Didn't know they offered this.

PR #136, opened 2017-01-23
use <project_root>/.babelrc file if exists; otherwise use default config

stelioskiayias:
The user defined .babelrc file was ignored. A string was returned instead of the file's content.


JeffreyWay:
It should automatically read any .babelrc file in your project root. I just checked, and it did.

stelioskiayias:
@JeffreyWay Yeah, you are right. I double checked. Something else must have went wrong. Sorry.

PR #149, opened 2017-01-24
Added validation for some public methods to provide better error reporting to users

jameshulse:
The current error that is received when omitting a required parameter is not very useful for a developer quickly trying to figure out where they are going wrong. Hopefully this change will help people realise their mistake.


JeffreyWay:
Thanks!

PR #166, opened 2017-01-25
source maps for less

SebastianS90:
`resolve-url-loader` is only for sass. It fails when using less-loader with source map, so I had to split that.


webmarkelov:
Alternatively there is a solution  -> **Support for less-loader**  https://github.com/bholloway/resolve-url-loader/pull/41

JeffreyWay:
Thanks!

Issue #184, opened 2017-01-26
Combine sass (or less)?


jpmurray:
Hello,

The docs speaks of the `mix.combine()` method, that allows to combine files. But it would seems that it only combine files that are already compiled (I might be wrong, if I am, sorry!).

With Elixir, I could combine multiple sass and less files in one big compiled css file like so:

```
mix.sass([
        'app.scss',
        'controllers.scss'
    ], 'public/assets/css');
```

Trying the same with laravel-mix results in an error `Path must be a string`, so that's clearely not working. Is it impossible at the moment to do so? Or did I miss something entirely?

JeffreyWay:
You should have your main entrypoint Sass file import any other files that it needs to be combined with.

jpmurray:
It... Totally makes sense. Thank you for that, I don't really know why I stuck with what I did before actually ! :D

Zedonboy:
me too i wasn't smart..
thanks @JeffreyWay you're my Hero

PR #195, opened 2017-01-26
Add ability to compile less and sass files

donnysim:
Adds ability to compile less and sass files in one go,

```
mix.less('resources/assets/less/style.less', 'public/assets/css/less.css')
   .less('resources/assets/less/style2.less', 'public/assets/css/less2.css')
   .sass('resources/assets/sass/style2.scss', 'public/assets/css')
   .sass('resources/assets/sass/style.scss', 'public/assets/css');
```

### DIFF
--- a/setup/webpack.config.js
+++ b/setup/webpack.config.js
@@ -132,7 +132,7 @@ if (Mix.stylePreprocessors) {
                         'css-loader',
                         'postcss-loader',
                         'resolve-url-loader',
-                        (type == 'sass') ? 'sass-loader?sourceMap' : 'less-loader'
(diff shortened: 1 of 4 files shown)


renatomjunior22:
I need this in one go too! I write my styles in less, but I need bootstrap 12/24-col grid system! (Which is in sass/scss)

patie:
possible solution https://github.com/JeffreyWay/laravel-mix/issues/184#issuecomment-275385555

JeffreyWay:
Thanks for the PR! There were a few naming things here that didn't gel well, so I pushed a commit to add support just now. It'll be tagged later today.

donnysim:
No problem, now I can compile both and it's all that matters :)

PR #209, opened 2017-01-27
fix chunk loading

### LINKED ISSUE #164: Webpack dynamic chunk loading not working because of publicPath setting
I have a single page application and use `vue-router` in combination with async components. Webpack therefore creates separate chunks for each page and they are loaded when the user wants to access the route.

Dynamic chunk loading is not working right now because of the `output.publicPath` setting. In the default Mix configuration, webpack `output.publicPath` [is set](https://github.com/JeffreyWay/laravel-mix/blob/55d07f6381a0de5e116b3eaefb92de41548d55de/src/Mix.js#L109) to `./` and javascript files have the path `/js/[name].js`.

When webpack wants to load a new chunk, then it inserts a script tag:
```javascript
script.src = __webpack_require__.p + "/js/" + chunkId + ".js";
```
where `__webpack_require__.p = "./"` because of the mentioned `publicPath` setting.

For chunk id 7, the resulting path is `.//js/7.js` and will be expanded to `http://localhost:8000//js/7.js`.
But the `php artisan serve` webserver does not deliver the `public/js/7.js` file because of the additional `/` in the url. Instead, we get a 404 error.

Even worse, when we are on a `vue-router` route `user/17/edit` and webpack wants to load chunk 7, then the resulting url is `http://localhost:8000/user/17/edit/js/7.js` and it will obviously not work.

We cannot just change `publicPath` to `/` (without `.`), because then we will get `//js/7.js` which is in fact `http://js/7.js` and makes no sense either.

The cleanest solution probably is the following:
- Webpack's `output.publicPath` will be set to `/` in normal operation. When using HMR, then the path should be `http://localhost:8080/`. This can be achieved by modifying [this line](https://github.com/JeffreyWay/laravel-mix/blob/55d07f6381a0de5e116b3eaefb92de41548d55de/src/Mix.js#L109).
- Script names should not start with a leading slash, i.e. the names are `js/app.js`, `js/7.js`, ...
This can be achieved by modifying [this line](https://github.com/JeffreyWay/laravel-mix/blob/55d07f6381a0de5e116b3eaefb92de41548d55de/src/Mix.js#L108) to `.replace(this.publicPath + '/', '')`.
- The `Manifest.transform` function will ensure that all paths in the `mix-manifest.json` file will start with a slash as required by laravels `mix` helper function.
This can be achieved by inserting `if (!path.startsWith('/')) path = '/' + path;` above [this line](https://github.com/JeffreyWay/laravel-mix/blob/55d07f6381a0de5e116b3eaefb92de41548d55de/src/Manifest.js#L24).

Please let me know what you think of my proposed solution. If you agree, then I'll create a PR.

SebastianS90:
closes #164

### DIFF
--- a/src/Mix.js
+++ b/src/Mix.js
@@ -87,11 +87,13 @@ class Mix {
      */
     output() {
         let filename = this.versioning ? '[name].[chunkhash].js' : '[name].js';
+        let chunkFilename = this.js.base + '/' + (this.versioning ? '[name].[chunkhash].js' : '[name].js');
 
         return {
             path: this.hmr ? '/' : this.publicPath,
             filename: filename,
-            publicPath: this.hmr ? 'http://localhost:8080/' : './'
+            chunkFilename: chunkFilename,
+            publicPath: this.hmr ? 'http://localhost:8080' : ''
         };
     }
 


SebastianS90:
This makes [my example](https://github.com/SebastianS90/laravel-mix-vue-router) work.

The resulting manifest contains:
```javascript
__webpack_require__.p = "http://localhost:8080"; // for HMR
__webpack_require__.p = ""; // otherwise

script.src = __webpack_require__.p + "/js/" + ({"0":"foo","1":"foobar","2":"/js/vendor","3":"/js/app"}[chunkId]||chunkId) + ".js";
```

It feels a bit strange that vendor and app are listed with `/js/` twice in the URL, but that shouldn't matter because they aren't loaded dynamically (you have to reference them in the html file)

laugharn:
This would be very helpful.

jordonbaade:
Hoping this is tagged for release soon, I'm just overwriting Mix.js source until it's safe to update.  Thanks @SebastianS90, @JeffreyWay!

PR #235, opened 2017-01-30
Docs: Update script names to match with Laravel's package.json

adriaanzon:
I updated only this section because this one is specifically about Laravel.

However there are a few other sections with different script names, but they aren't specifically about Laravel ([installation](https://github.com/JeffreyWay/laravel-mix/blob/master/docs/installation.md#npm-scripts) and [faq](https://github.com/JeffreyWay/laravel-mix/blob/master/docs/faq.md#my-code-isnt-being-minified)). Shall I update those as well?


adriaanzon:
Well, shall I? :smile:

Issue #63, opened 2017-01-07
HMR not Working when using Copy Plugin


SeoFood:
npm run hmr
```javascript
mix.js('resources/assets/js/app.js', 'public/js')
mix.copy('node_modules/bootstrap/fonts', 'public/fonts')
```

This error preventing hot reload from working:
```
ERROR in [copy-webpack-plugin] Using older versions of webpack-dev-server, devServer.outputPath must be defined to write to absolute paths
```

npm run hmr
```javascript
mix.js('resources/assets/js/app.js', 'public/js')
//mix.copy('node_modules/bootstrap/fonts', 'public/fonts')
```

Works.

I tried to use the output path as an absolute path but it can't find any js file.

QWp6t:
Try:
```sh
yarn add webpack-dev-server@2.2.0-rc.0 --dev
# npm i webpack-dev-server@2.2.0-rc.0 -D
```

SeoFood:
yeah tried this already. It is related to the copy-webpack-plugin

https://github.com/kevlened/copy-webpack-plugin/issues/29

jbreuer95:
@SeoFood Any workaround? Downgrading copy-webpack didnt work for me

SeoFood:
@jbreuer95 no not yet sorry, spend an hour to find the problem. The copy plugin doesn't seem to be really stable. Dunno if there are better webpack copy plugins who support the webpack dev server.

nmfzone:
So, until this fixed, we can't use `hmr`? Because I have `.copy()` in my `webpack.mix`. Sad :/

SeoFood:
Yes didn't find a fix for that yet.

jbreuer95:
@SeoFood thanks for trying :)

solicomo:
It seems to have been fixed on master. #150 

waiting for new release...

jbreuer95:
@solicomo 
it still doesn't work for me, for now i remove the mix copy's when i run hmr and user ../ paths instead of / paths.

rmartins90:
Any solution for this?

jbruni:
My temporary solution is to wrap the `mix.copy` entries in a condition, as follows:

```
if (mix.config.inProduction) {
    mix.copy('node_modules/bootstrap/fonts', 'public/fonts');
}
```

Of course, to have the assets copied, I need to use `npm run production` once.

JeffreyWay:
This is fixed in the next tagged release.

[...]

PR #239, opened 2017-01-30
Fix dev-server error when using copy. fix for issue #63

shemi:
When I used the copy function and run npm run dev I get the following error:

ERROR in [copy-webpack-plugin] Using older versions of webpack-dev-server, devServer.outputPath must be defined to write to absolute paths

*On Windows
**Not in leravel project

#63

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -148,7 +148,8 @@ module.exports.combine = (src, output) => {
 module.exports.copy = (from, to, flatten = true) => {
     Mix.copy = (Mix.copy || []).concat({
         from,
-        to: Mix.Paths.root(to),
+        to,
+        context: Mix.Paths.root(),
         flatten: flatten
     });
 


JeffreyWay:
Hmm I'm not getting that error on this end. Let me double check if we can get away with removing the base path.

betteryourweb:
I'm getting the same error. I only get the error when I try to run npm dev hot

Distributor ID:	Debian
Description:	Debian GNU/Linux 8.6 (jessie)
Release:	8.6
Codename:	jessie

bloggervista:
i also got same error on windows :(

JeffreyWay:
This is a breaking change. If I apply your PR and do:

```
mix.copy('foo.txt', 'public')
```

It puts the file in "public/public/foo.txt".

alairock:
Having this issue in laravel spark. Spark is using an older version of mix, and that might be the problem. 

Regardless. If I comment out the `mix.copy` lines, HMR works again. 

Hope this helps someone else. Could be useful for someone to go make a PR over at the Spark repo too. I don't have the JavaScript chops to do that, I'm afraid.

PR #279, opened 2017-02-03
set sass-loader sourceMap in options array

forehalo:
Instead of setting in query parameters which causes confusing.


JeffreyWay:
This is because source maps are required for CSS extraction to work. So I don't want users to think that they can override that setting.

forehalo:
@JeffreyWay  but, under my envirnment, using `?sourceMap` will always cause this error:

image

just run the laravel-mix-example, nothing changed.

PR #288, opened 2017-02-03
Fix record function in Version class returning this

mayeco:
This simple "return this" makes the following code works when the manifest does not exist.

```
            this.versioning = new Versioning(this.manifest).record();

            this.events.listen('build', () => {
                this.versioning.prune(this.publicPath);
            });
```

```
this.versioning.prune(this.publicPath);
                               ^
TypeError: Cannot read property 'prune' of undefined

```

### DIFF
--- a/src/Versioning.js
+++ b/src/Versioning.js
@@ -19,7 +19,7 @@ class Versioning {
      * Record versioned files.
      */
     record() {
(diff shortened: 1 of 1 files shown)


mayeco:
ping @JeffreyWay

Issue #270, opened 2017-02-02
make Version() work on combine() and copy() plz


reno1979:
Related to: 
https://github.com/JeffreyWay/laravel-mix/issues/53
https://github.com/JeffreyWay/laravel-mix/issues/154
https://github.com/JeffreyWay/laravel-mix/issues/163
https://github.com/JeffreyWay/laravel-mix/issues/211

@JeffreyWay Perhaps you will consider a possible fix or workaround for this. 

We combine a bunch of vendor files (allready minified etc) , but sometimes we want to change the content and make sure the visitors get the updated combined file, the version() helper would be very important for us here.

hovsep:
"minify" and other transformers not work on collections after copy\combine etc

huglester:
hello,

this one would be lovely.
Because even first combining and trying to version combined files does not work...

```bash
mix.combine([
    'addons/default/webas/default-theme/resources/vendor/js/vendor/jquery.min.js',
    'addons/default/webas/default-theme/resources/vendor/js/vendor/what-input.min.js',
    'addons/default/webas/default-theme/resources/vendor/js/foundation.min.js',
    'addons/default/webas/default-theme/resources/vendor/js/motion-ui.min.js',
    'addons/default/webas/default-theme/resources/vendor/js/blueimp-gallery.min.js',
    'bower_components/typeahead.js/dist/typeahead.bundle.js',
    'bower_components/handlebars/handlebars.js',
    'addons/default/webas/default-theme/resources/js/typeahead.js',
], 'public/assets/js/app.js');

mix.js('public/assets/js/app.js', 'public/assets/build/js/app.js')
    .version();
```


```bash
npm ERR! Darwin 16.4.0
npm ERR! argv "/usr/local/Cellar/node/7.5.0/bin/node" "/usr/local/bin/npm" "run" "dev"
npm ERR! node v7.5.0
npm ERR! npm  v4.1.2
npm ERR! code ELIFECYCLE
npm ERR! @ dev: `node node_modules/cross-env/bin/cross-env.js NODE_ENV=development node_modules/webpack/bin/webpack.js --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js`
npm ERR! Exit status 2
npm ERR!
[...]

PR #297, opened 2017-02-04
make combined files working with version! issue #270

forehalo:
image
image

[#270](https://github.com/JeffreyWay/laravel-mix/issues/270)

### DIFF
--- a/src/Concat.js
+++ b/src/Concat.js
@@ -1,5 +1,7 @@
 let chokidar = require('chokidar');
 let concatenate = require('concatenate');
+let mkdir = require('mkdirp');
+let md5 = require('md5');
 let File = require('./File');
 
 class Concat {
@@ -9,19 +11,33 @@ class Concat {
      * @param {object} events
      */
     constructor(events) {
-        this.files = [];
-
-        events.listen('build', this.run.bind(this));
+        this.combinations = [];
+        this.options = null;
+        this.events = events;
     }
 
+    /**
+     * Initialize the concatenate.
+     * 
+     * @param  {object} options
+     */
+    initialize(options) {
+        this.options = options;
+        this.events.listen('build', this.run.bind(this));
+        this.watch();
+    }
 
     /**
      * Add a set of files to be concatenated.
      *
      * @param {object} files
      */
-    add(files) {
-        this.files = this.files.concat(files);
+    add(combination) {
+        this.combinations = this.combinations.concat({
+            src: combination.src,
+            output: combination.output,
+            originOutput: combination.output
+        });
     }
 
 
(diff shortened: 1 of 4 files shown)


JeffreyWay:
Thanks for this. One thing I noticed is that you're passing an array of files to `mix.version()`. But your PR doesn't support that, right?

forehalo:
@JeffreyWay Uh, the array passed was not used. We should still use `mix.version()` to generate versioned files. It was to show they can be versioned successfully, nothing else.

JeffreyWay:
Okay this is merged in now. Thanks!

iraklisg:
@forehalo thank you so much for this helpful addition!  @JeffreyWay the current npm version of laravel-mix@6.0.3 does not support this feature yet. Are there any plans to be included in laravel-mix@6.0.4

JeffreyWay:
Yes - it'll be tagged once I do some more testing.

nejtr0n:
it still not combine compiled version css. For example
```
.sass('admin/core.scss', 'public/css/admin/core.css')
.combine([
'public/css/admin/extra.css',
'public/css/admin/core.css'
], 'public/css/admin/all.css'
).version()
```
I've got an error - file not founded.
(because public/css/admin/core.css has versioned name like core.asdase242.css)

iraklisg:
@nejtr0n the same here, the PR has not yet been accepted so patience :stuck_out_tongue: However, until the issue is solved, I think that @JeffreyWay should consider changing the [docs](https://github.com/JeffreyWay/laravel-mix/commit/27902ec2f6720e39156e9ed1b111c15fbfc634db#diff-bb1bd93b5077affc1930c467e3197e96R55) where it is implied that versionning applies  to combined files too

rjvim:
Is this fix available in "laravel-mix": "^0.8.1", Sorry It's not clear if this fix is already available as I am facing the same issue.

JeffreyWay:
@iraklisg The PR has been accepted, so no changes to the docs are needed.

https://www.dropbox.com/s/jrz4ri46k5cr2sv/Screenshot%202017-02-19%2015.23.48.png?dl=0

JeffreyWay:
@nejtr0n You're getting an error because `public/css/admin/core.css` doesn't exist. You requested versioning, so it create the hashed version of that file. Then, you're trying to combine that file, but, again, because it doesn't exist, it fails. 

Instead, use a regular expression.

```js
.sass('admin/core.scss', 'public/css/admin/core.css')
.combine([
    'public/css/admin/extra.css',
    'public/css/admin/core.*.css'
], 'public/css/admin/all.css')
.version()
```

nejtr0n:
Thank you! Didn't know about ability of regexp in combine =)

Issue #305, opened 2017-02-05
assets with relative url paths are not resolved due to webpack config bug (>v0.6.0)


parapilot:
In line 142ff in node_modules/laravel-mix/setup/webpack.config.js there is

```
                        {
                            loader: 'sass-loader?sourceMap',
                            options: Object.assign({
                                precision: 8,
                                outputStyle: 'expanded'
                            }, toCompile.pluginOptions)
                        }
                    ] : [
                        {
                            loader: 'less-loader' + sourceMap,
                            options: toCompile.pluginOptions
                        }
```

this cannot work because the sourceMap option is still passed within the loader string but then lost because an option object is passed. To work again this should be something like:

```
         {
                            loader: 'sass-loader',
                            options: Object.assign({
                                sourceMap: true,
                                precision: 8,
                                outputStyle: 'expanded'
                            }, toCompile.pluginOptions)
                        }
                    ] : [
                        {
			    loader: 'less-loader',
			    options: Object.assign({
                                sourceMap: true
                            }, toCompile.pluginOptions)
                        }
```

In the same section there are also the deprecated parts that causes warnings:

```
        module.exports.module.rules.push({
            test: new RegExp(toCompile.src.path.replace(/\\/g, '\\\\') + '$'),
            loader: extractPlugin.extract({
                fallbackLoader: 'style-loader',
                loader: [
                    { loader: 'css-loader' + sourceMap },
                    { loader: 'postcss-loader' + sourceMap }
which should be updated to:
module.exports.module.rules.push({
            test: new RegExp(toCompile.src.path.replace(/\\/g, '\\\\') + '$'),
            use: extractPlugin.extract({
                fallback: 'style-loader',
                use: [
                    { loader: 'css-loader' + sourceMap },
                    { loader: 'postcss-loader' + sourceMap }
```

Sincerely yours
parapilot

forehalo:
Got the same error. #279 

JeffreyWay rejected this PR because the `sourceMap` of sass-loader is strictly required, should not be override by user's config.

So, you can change the sass-loader to:
```
{
    loader: 'sass-loader',
        options: Object.assign({
            precision: 8,
            outputStyle: 'expanded'
    }, toCompile.pluginOptions, { sourceMap: true })
}
```

I'll create a PR soon.

JeffreyWay:
We fixed this in a recent PR.

PR #312, opened 2017-02-06
fix sass-loader issue

forehalo:
#305 

I've create a PR #279 before, but rejected because the `sourceMap` of sass-loader is required and should not be override by user setting.

Seems that use both `query parameters` and `loader options` still get conflict and error.

So, pass `sourceMap` setting in the last option to override all set before.

### DIFF
--- a/setup/webpack.config.js
+++ b/setup/webpack.config.js
@@ -140,11 +140,11 @@ if (Mix.preprocessors) {
                     toCompile.type == 'sass' ? [
                         { loader: 'resolve-url-loader' + sourceMap },
                         {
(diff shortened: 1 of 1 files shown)


19peaches:
This merge didn't fix issue #279, fonts are still not found in versions greater than 0.6

JeffreyWay:
We haven't tagged this commit yet.

19peaches:
ah ... ok thanks @JeffreyWay

nmfzone:
Waiting on this.

earllapura:
+1

PR #318, opened 2017-02-07
extract vue single-file-component styles

forehalo:
usage: `mix.extractVueStyle('public/css/vue.css')`

An simple version need cleaned up according to [vue-cli](https://github.com/vuejs-templates/webpack/blob/master/template/build/utils.js#L12).

### DIFF
--- a/setup/webpack.config.js
+++ b/setup/webpack.config.js
@@ -74,22 +74,6 @@ module.exports.output = Mix.output();
 module.exports.module = {
     rules: [
         {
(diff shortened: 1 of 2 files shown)


adriaanzon:
Would it be possible to make this a bit more generic and extract all styles that are imported in JavaScript?

JeffreyWay:
This is already possible with:

```js
mix.options({
    extractVueStyles: true
});
```

PR #327, opened 2017-02-07
Ability to set prefix for generated asset paths, fixes #339

### LINKED ISSUE #339: [help] How to set a different base path for images?
Hi,

I'm using the package as stand-alone and I need to change the base path for sass images.
I now I can change the `webpack.config.js` file but, I'm wondering if there's a better way to do that?
Maybe using the `mix.webpackConfig()` method? Otherwise, every time that I update the packaged I need to compare the package `webpack.config.js` against mine and merge the changes, I know, is not that hard but, if there's any way to save time, I'd prefer using that.
Thanks!

kmorva:
Currently the `publicPath` of image and font resources is hard-coded to `/`. However, it's quite usual, that someone wants the website to work even from a sub-folder, for example `localhost/mypage`. Laravel provides an `asset` helper for it, which generates a fully qualified URL, but because of this setting in Laravel Mix, it becomes meaningless, because resources linked from CSS files won't work anyways.
I propose a solution for giving the ability to the developer to set the resource root used for generated resource locators with `setResourceRoot`. This way, developers don't have to hold a copy of `webpack.config.js` if they want to support sub-folder websites.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,22 @@
+<!-- *Before creating an issue please make sure you are using the latest version of laravel-mix.* -->
+
+### What is the current behavior?
+
+
+
+### What is the expected or desired behavior?
+
+
+
+---
+
+**Steps to reproduce, including full log output:**
+<!-- If you can, provide a link to a public repository which contains the files necessary to reproduce this. -->
+
+
+
+**Local environment:**
+
+Operating System: X
+
+NPM/Node version: X
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,22 +1,24 @@
-<!-- *Before creating an issue please make sure you are using the latest version of laravel-mix.* -->
+<!-- 
+    Before posting this issue, please perform the following steps, and confirm that the issue is still present.
 
-### What is the current behavior?
+1. Update Laravel Mix to the latest version.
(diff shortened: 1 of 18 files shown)


adenvt:
i install this branch and got error:

```
Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
 - configuration.module.rules[5].use should be one of these:
   non-empty string | function | object { loader?, options?, query? } | function | [non-empty string | function | object { loader?, options?, query? }]
   Details:
    * configuration.module.rules[5].use should be a string.
    * configuration.module.rules[5].use should be an instance of function.
    * configuration.module.rules[5].use should be an object.
    * configuration.module.rules[5].use should be one of these:
      non-empty string | function | object { loader?, options?, query? }
    * configuration.module.rules[5].use should be an instance of function.
    * configuration.module.rules[5].use[1] should be a string.
    * configuration.module.rules[5].use[1] should be an instance of function.
    * configuration.module.rules[5].use[1] should be an object.
    * configuration.module.rules[5].use[1] should be one of these:
      non-empty string | function | object { loader?, options?, query? }
```

this my mix:
```
mix.setResourceRoot('/')
    .sass('resources/assets/sass/app.scss', 'public/css')
    .js('resources/assets/js/app.js', 'public/js')
    .version()
```

kmorva:
@adenvt are you sure the `master` branch working for you?
What if you don't call `setResourceRoot`, does it still break?

kmorva:
@adenvt I have rebased and tested, I don't experience your issue, I think it must be on your side.
I've created a new PR with the rebased commits (have no idea why did rebasing mess up the history of this PR): #353

PR #347, opened 2017-02-08
Add support for loading html files

### LINKED ISSUE #346: Issues with loading .html files
The current version of laravel mix does not support vue components that loads an external .html file as a template.

You will see errors like this if you try to compile with version 0.6.3 of laravel mix:
```
Module parse failed: .../resources/assets/js/Vue/Components/UserLoginModal/UserLoginModal.html Unexpected token (1:0)
You may need an appropriate loader to handle this file type.
```

ventrec:
Fixes #346

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "file-loader": "^0.9.0",
     "friendly-errors-webpack-plugin": "^1.1.3",
     "fs": "0.0.1-security",
+    "html-loader": "^0.4.4",
     "less": "^2.7.1",
     "less-loader": "^2.2.3",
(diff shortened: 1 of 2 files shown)


JeffreyWay:
Do we need an update to package.json that installs html-loader?

ventrec:
I updated package.json with the missing html-loader. Should be ok now i guess.

aromka:
Is there a way to either disable this or overwrite? 
I need to use my own html loader (minify html, etc), and this rule breaks the process (outputs double module.exports = "module.exports = ...".

PR #348, opened 2017-02-08
Fixing JSDoc

kmorva:
- Fixing wrong parameter names.
- `array` is not a type, `Array` is: http://usejsdoc.org/tags-type.html#examples
- Instead of `mixed`, `{*}` should be used: http://usejsdoc.org/tags-param.html#multiple-types-and-repeatable-parameters
- `bool` is not a type, `boolean` is: http://usejsdoc.org/tags-type.html#overview

### DIFF
--- a/src/Collection.js
+++ b/src/Collection.js
@@ -13,7 +13,7 @@ class Collection {
      * Add a new key-value pair to the collection.
      *
      * @param {string}       name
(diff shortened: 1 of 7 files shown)


JeffreyWay:
Thanks! Good description, too. 👍

PR #351, opened 2017-02-08
Resolves issue with special characters in dir

bradenkeith:
Proposing this as a fix for Issue 317 where spaces, ( and ) are not escaped appropriately. In addition, move from a regex expression to a normalize() command to get path structures functional cross platform. Requesting windows user support in confirming normalize() still allows file paths to properly escape

### DIFF
--- a/setup/webpack.config.js
+++ b/setup/webpack.config.js
@@ -149,7 +149,7 @@ if (Mix.preprocessors) {
         let sourceMap = Mix.sourcemaps ? '?sourceMap' : '';
 
         module.exports.module.rules.push({
-            test: new RegExp(toCompile.src.path.replace(/\\/g, '\\\\') + '$'),
+            test: new RegExp(toCompile.src.path.replace(/(\s|\(|\))/g,'\\$1').normalize() + '$'),
             use: extractPlugin.extract({
                 fallback: 'style-loader',
                 use: [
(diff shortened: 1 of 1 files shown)


bradenkeith:
-Does this work for Windows users?
-Are there any other special characters that should be considered for escaping? Currently supported: spaces, (, )

QWp6t:
I think you're a little confused about that normalize() method.

Looks like you think you're using [`path.normalize()`](https://nodejs.org/api/path.html#path_path_normalize_path), but you're actually using [`String.prototype.normalize()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize).

bradenkeith:
You're dang straight I'm a little confused! 😆 

This is my first node contribution, so I was not aware of the chaining abilities and yes it was meant to be called on the path, not the string.

According to commit notes, we don't need this due to the last commit covering Windows users. 

```
Normalize was being called on the string instead of the path. It was introduced to remove replace(/\\/g, ‘\\\\’), which we have determined is still required.
```

bradenkeith:
Hey @JeffreyWay - What's the next steps for getting this merged?

JeffreyWay:
I'm hesitant to merge this because I don't have any tests to confirm that it does in fact work.

bradenkeith:
The support thread includes a few different platforms confirming fix, but I'm more hesitant to merge because I feel this might be more of a webpack level issue and shouldn't be fixed at each plugin level. Shouldn't their file management portion be able to escape characters? They already do for some, just doesn't seem like it does a perfect job yet.

PR #353, opened 2017-02-09
Ability to set prefix for generated asset paths (Fixes #339)

### LINKED ISSUE #339: [help] How to set a different base path for images?

KristofMorva:
Original PR: #327

Currently the `publicPath` of image and font resources is hard-coded to `/`. However, it's quite usual, that someone wants the website to work even from a sub-folder, for example `localhost/mypage`. Laravel provides an `asset` helper for it, which generates a fully qualified URL, but because of this setting in Laravel Mix, it becomes meaningless, because resources linked from CSS files won't work anyways.
I propose a solution for giving the ability to the developer to set the resource root used for generated resource locators with `setResourceRoot`. This way, developers don't have to hold a copy of `webpack.config.js` if they want to support sub-folder websites.


dstollie:
Thanks for this pr, this fixed my issue!

irazasyed:
Need this merged 👍 

I was wondering why I see fonts and images directory in public root when I'm trying to copy it to assets dir. It does copy to assets directory but at the same time copies the files to public dir. That was quite confusing TBH.

Hopefully this gets merged.

PR #362, opened 2017-02-11
Fix broken test by adding assertion that will properly throw .. 

ericnkatz:
.. if paths in combine() params go to files that do not exist.

### DIFF
--- a/src/Verify.js
+++ b/src/Verify.js
@@ -64,6 +64,13 @@ class Verify {
             Array.isArray(src),
             'mix.combine() requires an array as its first parameter.'
         );
+
+        Array.from(src).map(file => {
+            assert(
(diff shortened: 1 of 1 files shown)


ericnkatz:
This PR just adds a loop that runs an assert() for each item passed into src array. It uses Array.from() to guarantee we use map() on an array instance (in case you change the order in which you call assertions).

I wasn't sure what messaging made sense for failure so if you would like me to change it let me know.

JeffreyWay:
I removed that assertion a few commits ago on purpose. So if there is a lingering test, we just need to delete it.

ericnkatz:
K, @JeffreyWay do you want a PR for that?

Issue #359, opened 2017-02-10
mix-entry.js missing on less compile


faustbrian:
- Laravel Mix Version: 0.7.4
- Node Version (`node -v`):  v7.5.0
- NPM Version (`npm -v`):  4.2.0
- OS: macOS Sierra 10.12.3

### Description:

Everytime I try to compile my less files this error comes up. I tried to compile sass files on another project and there it is working, seems to be an issue with less files.

Setup is fresh and without any changes to the webpack.mix.js besides compiling less instead of sass.

```
 ERROR  Failed to compile with 1 errors

 error  in ./mix-entry.js

Module build failed: Error: ENOENT: no such file or directory, open '/Users/Brian/Sites/project/mix-entry.js'

 @ multi ./mix-entry.js ./resources/assets/less/app.less
```

JeffreyWay:
Please paste in your webpack.mix.js file.

faustbrian:
```
const { mix } = require('laravel-mix');

/*
 |--------------------------------------------------------------------------
 | Mix Asset Management
 |--------------------------------------------------------------------------
 |
 | Mix provides a clean, fluent API for defining some Webpack build steps
 | for your Laravel application. By default, we are compiling the Sass
 | file for the application as well as bundling up all the JS files.
 |
 */

mix.js('resources/assets/js/app.js', 'public/js')
   .less('resources/assets/sass/app.less', 'public/css');
```

JeffreyWay:
Are you sure that's your Mix file? The error you're getting doesn't match up with that. 

Also, it's squawking about `./resources/assets/less/app.less`, but the Mix file you pasted has `'resources/assets/sass/app.less`

faustbrian:
Yeah I renamed the folder to less but this is my mix file.

JeffreyWay:
The thing is that `mix-entry.js` will only ever be created if you don't call `mix.js()`. So I'm not sure how it's possible that you're seeing this error with that Mix file. Hmm..

kuroisuna:
I also had this problem early in the morning:

```
 WAIT  Compiling...

 95% emitting
 ERROR  Failed to compile with 1 errors

 error  in ./mix-entry.js

Module build failed: Error: ENOENT: no such file or directory, open '/Users/axel/Development/misc/cmclux/mix-entry.js'
    at Error (native)

 @ multi ./mix-entry.js ./resources/assets/sass/app.scss

                            Asset       Size  Chunks             Chunk Names
[...]

PR #366, opened 2017-02-11
add a mock entry for no mix.js() config

forehalo:
#359 
This error was because of removing `mix-entry.js`  immediately after the first building in watch mode, so the next building after source code changed get crashed if no `mix.js()` called.

Add a `mock-entry.js` file in laravel-mix src for this scenario, instead of generating a temporary entry file and delete it after built.


JeffreyWay:
Thanks!

PR #418, opened 2017-02-18
Updated FAQs to make --watch-poll instructions clearer

JackWH:
At a quick glance it implies you need to use `--watch-poll` *instead* of `--watch`, but you actually need to use them together, so I made that clearer for anyone else who (like me) missed it the first time.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -26,7 +26,7 @@ It's highly recommended that you add the following NPM scripts to your `package.
 
 ### I'm using a VM, and Webpack isn't picking up my file changes.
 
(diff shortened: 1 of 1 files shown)


QWp6t:
Somewhat related: that FAQ should maybe be updated to note that safe writing (or backupcopy if you're using vim) can also cause issues with webpack picking up file changes.

**InteliJ**
[Settings ▶︎ Appearance & Behavior ▶︎ System Settings ▶︎ Synchronization ▶︎ `safe write` (uncheck)](https://www.jetbrains.com/help/phpstorm/2016.3/system-settings.html#d792972e131)

**vim**
[`:set backupcopy=yes`](http://vimdoc.sourceforge.net/htmldoc/options.html#'backupcopy')

Issue #413, opened 2017-02-17
CSS injection using browsersync?


marcamos:
- Laravel Mix Version: 0.8.1
- Node Version (`node -v`): 6.9.1
- NPM Version (`npm -v`): 4.0.1
- OS: OS X 10.11.6

### Description:

I'm a front-end developer who's recently been interested in moving from gulp to webpack, but trying to learn webpack always leaves me confused. Laravel Mix seems like it'll help me make the leap.

The first order of business, for me, was to ensure BrowserSync works the way I'm used to it working over in gulp. I was super-happy to see support for it appear in [0.7.4](https://github.com/JeffreyWay/laravel-mix/releases/tag/v0.7.4) and immediately tried it out. I got it up and running using the following:

```
mix.js('_js/app.js', 'html/lib/js')
  .sass('_scss/app.scss', 'html/lib/css')
  .sourceMaps()
  .browserSync({
    proxy: 'http://starter.vbox.bytheoutfit.com/',
    files: [
      'html/_demos/**/*.+(html|php)',
      'mix-manifest.json',
      'html/lib/css/**/*.css',
      'html/lib/js/**/*.js'
    ]
  });
```
…where `http://starter.vbox.bytheoutfit.com/` points to my local VirtualBox.

As expected, the watch script refreshed all connected browsers, which is great, but css-only file-changes resulted in an entire page-refresh as well, whereas I was hoping the compiled css would be injected.

It's not a huge deal, but I got very used to scss/css updates being injected and wonder if that's possible here, using Mix/webpack?

Thanks for your hard work, and for your time!

JeffreyWay:
Fixed in #430

marcamos:
Just wanted to write in and say that this now works perfectly. Thanks so much, @JeffreyWay.

pcampina:
IN 2020, config is:

```
mix
  .js("resources/js/app.js", "public/js")
  .sass("resources/sass/app.scss", "public/css")
  .browserSync({
    proxy: 'localhost', // your domain test
    port: 32816 // your port
  });
```

PR #430, opened 2017-02-20
Allow for BrowserSync to inject css files without reloading the page.

SethTompkins:
This would close issue #413 - the Webpack BrowserSync plugin accepts a second argument that is an object of options for the plugin itself, not for BrowserSync. The important option here is `reload`, if it is set to false the plugin allows for injection rather than hard reloading.

It is also important to remove the matcher for mix-manifest.json. Even with reload set to false, files listed in `files` that change and are not deemed injectable will trigger a hard reload.

### DIFF
--- a/setup/webpack.config.js
+++ b/setup/webpack.config.js
@@ -331,18 +331,22 @@ module.exports.plugins = (module.exports.plugins || []).concat([
 
 if (Mix.browserSync) {
     module.exports.plugins.push(
-        new plugins.BrowserSyncPlugin(Object.assign({
-            host: 'localhost',
-            port: 3000,
-            proxy: 'app.dev',
-            files: [
-                'app/**/*.php',
-                'resources/views/**/*.php',
-                'public/mix-manifest.json',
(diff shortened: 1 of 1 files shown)


JeffreyWay:
Thanks!

giovannipds:
Weird, I thought that the files .css should be on the property, not the pre-processor ones. With .css it wasn't streaming with browserSync. Thanks for the tip, @sagalbot.

pixiwooh:
I cannot get it to work, even with files specified, it reloads the page at every scss change instead of just injecting them.

`mix.browserSync({
        files: [
            'src/**/*',
            'dist/assets/**/*',
            'dist/images/**/*',
            'dist/system/user/templates/**/*'
        ],
        proxy: {
            target: 'https://mydevwebsite.localhost/',
        },
        open: false
    });`

Where src is where my sass files are, dist/assets is where my compiled js files are, dist/system/user/templates is where my html files are (I only want reload for my html files and simple injects for the other types).

Anyone have an idea please?

PR #433, opened 2017-02-21
Replaced env() with config() in a code sample for LiveReload doc

kkomelin:
... because env() is not available when the config is cached.

Used the following command to cache the config:
```
php artisan config:cache
```


kkomelin:
I'll resolve conflicts.

PR #434, opened 2017-02-21
Replaced env() with config() in a code sample of Livereload doc

kkomelin:
... because env() is not available when the config is cached.


kkomelin:
Thanks @JeffreyWay

PR #464, opened 2017-02-24
Inject environment variables into scripts

adiachenko:
See #465.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "copy-webpack-plugin": "^4.0.1",
     "cross-env": "^3.1.3",
     "css-loader": "^0.14.5",
+    "dotenv": "^4.0.0",
     "extract-text-webpack-plugin": "^2.0.0-rc.3",
     "file-loader": "^0.9.0",
     "friendly-errors-webpack-plugin": "^1.5.0",
--- a/setup/webpack.config.js
+++ b/setup/webpack.config.js
@@ -3,6 +3,20 @@ let glob = require('glob');
 let webpack = require('webpack');
 let Mix = require('laravel-mix').config;
 let webpackPlugins = require('laravel-mix').plugins;
+let dotenv = require('dotenv')
+
+/*
+ |--------------------------------------------------------------------------
+ | Load environment variables
+ |--------------------------------------------------------------------------
+ |
+ | Load environment variables from .env file. dotenv will never modify
+ | any environment variables that have already been set.
+ |
+ */
+dotenv.config({
+    path: Mix.Paths.root('.env')
+});
 
 /*
  |--------------------------------------------------------------------------
@@ -376,17 +390,19 @@ if (Mix.options.purifyCss) {
 
 if (Mix.inProduction) {
     plugins.push(
-        new webpack.DefinePlugin({
-            'process.env': {
-                NODE_ENV: '"production"'
-            }
-        }),
-
         new webpack.optimize.UglifyJsPlugin(Mix.options.uglify)
     );
 }
 
 plugins.push(
+    new webpack.DefinePlugin(
+        Mix.definitions({
+            NODE_ENV: Mix.inProduction
+                ? 'production'
+                : ( process.env.NODE_ENV || 'development' )
+        })
+    ),
+
     new webpackPlugins.WebpackOnBuildPlugin(
         stats => Mix.events.fire('build', stats)
     )
--- a/src/Mix.js
+++ b/src/Mix.js
@@ -120,6 +120,33 @@ class Mix {
         return '?' + JSON.stringify(this.options.babel);
     }
 
+    /**
+     * Fetch definitions for DefinePlugin
+     *
+     * @param {object} merge
+     */
+    definitions(merge = {}) {
+        let regex = /^MIX_/i
+
+        // Filter out environment variables that doesn't pass regex
+        let env = Object.keys(process.env)
+            .filter(key => regex.test(key))
+            .reduce((value, key) => {
+                value[key] = process.env[key]
+                return value
+            }, {});
+
+        let values = Object.assign(env, merge);
+
+        return {
+            'process.env': Object.keys(values)
+                // Stringify all values so we can feed into Webpack DefinePlugin
+                .reduce((value, key) => {
+                    value[key] = JSON.stringify(values[key])
+                    return value
+                }, {})
+        };
+    }
 
     /**
      * Determine if we are working with a Laravel project.


QWp6t:
This is being misused here in my opinion, especially for `NODE_ENV`, which is deliberately changed depending on the type of environment you're _building_ for.

This would make sense if you're building your assets server-side. So locally you'd use .env and set NODE_ENV to development, and then remotely you'd use it to set NODE_ENV to production.

But this is a bad use case because it requires users to manually open their .env file and change the env depending on what kind of build they're doing. Whereas current workflow requires only that users specify it implicitly in the command line depending on the script they run via npm, i.e., `npm run dev` or `npm run production`.

adiachenko:
@QWp6t I think you're misunderstanding something here. 

It works exactly the same way as it did before for existing `laravel-mix` users with the value of `NODE_ENV` being set to `production` when running `npm run production`.

The only difference now lies in the fact how `NODE_ENV` variable is treated in other environments. Previously, DefinePlugin was enabled only in production and therefore `NODE_ENV` was resolved to `undefined` in other environments. Now, because we utilize DefinePlugin in a more proactive manner, it doesn't make sense to explicitly omit `NODE_ENV` for other builds. Therefore, we either get it's value from environment or default it to `development`. Again, this will still be overriden by `Mix.inProduction` check to avoid clashes with previous behavior.

Also, please note, that we load values from environment. That includes BOTH `.env` file AND current user's environment variables.

As I've already noted in the related issue, this is a [proven pattern](https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/config/env.js) that has been used in the wild for years and is very familiar to front-end engineers.

QWp6t:
I don't dispute the benefits of `.env` in the slightest, and I see no issues with adding dotenv to laravel-mix for those who wish to use it. I still question its usefulness in a build tool for a wider audience. API keys and other environment-specific information makes `.env` ideal at runtime, but at compile-time, I just don't see why it would be any more useful than simply defining the variables in your compile script for the particular environment you're targeting. But whatever. I can see how there are certainly cases, especially when laravel-mix is used outside of a Laravel project, where dotenv would be quite useful.

At this point I have to concede that I had totally missed your ternary operation, i.e., the part where you're still leaving in the old functionality making dotenv completely optional. I had just seen the giant red block in the diff and was like "lol y he do dat?" and probably should have looked at it a little more carefully. My initial impression was that you were attempting to replace the current workflow by requiring that NODE_ENV be set within .env.

### tl;dr

My initial gripes were unfounded. I assumed this was asking devs to define NODE_ENV in .env, but that's not the case at all. This PR adds dotenv, but its use is completely optional.

JeffreyWay:
I'm not against merging this, but my only concern is that very very few Mix users have a need for this. If others chime in here requesting that we merge this, I will.

seansch:
Just came across this PR looking to do this exact thing, has my vote.

jwalton512:
Outside of what is being offered in this PR, how do others handle managing variables between environments?  

The example I currently have (which googling bought me here) is storing the secret for Passport in my dev environment.  I would need to somehow modify that once in production.

QWp6t:
@jwalton512 You can use something like Ansible. Secret information can be encrypted with Ansible Vault.

There are other solutions, but [that's what we use](https://roots.io/trellis/docs/vault/).

adiachenko:
@jwalton512 You probably shouldn't use this feature for Passport `client_secret` because it's not intended to be shown to end users. DefinePlugin simply replaces one string (i.e. process.env.NODE_ENV) with the other ("production"). It doesn't automatically make it secure.

As a workaround, you can create separate route for your first party SPA (or whatever you have there on the receiving end) that dispatches internal request with attached client credentials to actual `/oauth/` uri. This will effectively make your client unaware of Oauth2 implementation behind it, and you could simply pass username and password like you would with a normal authentication flow. This is a bit naive, but probably still better that exposing private client information to your users.

MuraevArt:
I think this might be useful.

jwalton512:
@adiachenko that makes much more sense than what I was doing.  Thanks for the insight!

laugharn:
@JeffreyWay I think this is a really useful commit that tracks pretty well with the greater Laravel ecosystem: we're already leveraging code splitting in Mix, .env in Framework, and deployment scripts in Forge. Using Blade to echo variables into inline scripts has always felt a little gross, and as situations like component nesting where those echoes don't work become more common, this commit feels like it's skating to where the puck is probably going to be.

To your point of "very very few Mix users hav[ing] a need for this" I think a lot of users, myself included, simply weren't aware that this was a possibility. But I definitely feel like I need it now that I know that it is!

QWp6t:
I don't see how it tracks pretty well with the greater Laravel ecosystem, and I'm thinking that a few of you might be misunderstanding what this PR accomplishes. It's not going to have your client-side scripts reading from .env files. It's going to be leveraging .env during compilation and building static assets based on those variables.

[As noted in the repo this was poached from](https://github.com/facebookincubator/create-react-app/tree/master/packages/react-scripts/template#adding-custom-environment-variables):
> **The environment variables are embedded during the build time.** Since Create React App produces a static HTML/CSS/JS bundle, it can’t possibly read them at runtime.

This is not a typical use case for .env. The typical use case for .env is storing sensitive information that you don't want to commit to version control, and then reading from it at runtime. The [dotenv repo cites a few examples of how it should be used](https://github.com/motdotla/dotenv#should-i-commit-my-env-file), none of which should ever be exposed client-side, i.e., compiled into static assets that are served to clients via a build tool like Mix:
> It should only include environment-specific values such as database passwords or API keys. Your production database should have a different password than your development database.

This is why it's not actually useful to very many users of Mix, because Laravel  is consuming those variables. Most users of Mix aren't running node.js on their server and reading from .env at runtime. And if you're just compiling the variables, then they should be part of your build script and committed to version control to ensure that deployments are always consistent (otherwise a change to one of the variables requires changing uncommitted .env files at every point of deployment).

I guess I would challenge you guys to answer one question: **what problem is this actually solving?**

adiachenko:
> otherwise a change to one of the variables requires changing uncommitted .env files at every point of deployment

@QWp6t You're assuming most users of Mix are committing their compiled assets to version control (along with these injected variables) which is not exactly the case, despite this being the default in Laravel installation. In fact, even [laravel.com  doesn't follow this convention](https://github.com/laravel/laravel.com/blob/master/.gitignore). A lot of front-end engineers will frown on this and having worked with several Laravel projects, I've seen most people explicitly gone out of their way to ensure compiled stuff was out of version control. This isn't even the issue that should be addressed by abusing VSC. It's the responsibility of the CI and your devops if you don't want to build your scripts in production.

Anyway, I think there is still some misunderstanding here. Heck, most Laravel users haven't even worked with Webpack before. 3 to 4 months in we still don't have async chunk loading working or, in fact, any people concerned about it and that is the core webpack feature that would have immensely help in optimizing build size without introducing bizarre constructs to your layouts and mix config. That goes a long way to show that people need to be told about certain aspects of new tool they got to work with.

@laugharn actually gets the point of this PR. Setting environment variables via inline scripts in blade (like it's currently done) is extremely awkward.

Firstly, it's a multiple step process:
1. Create .env variable
2. Create config variable (because `env()` [doesn't get called with config caching in place](https://laravel.com/docs/5.4/configuration#configuration-caching))
3. Echo the variable in each concerned layout

Secondly, it's not really semantic. Laravel already utilizes `.env` file for its variables and you'll see Docker doing the same (e.g. ports, log settings etc.) in dockerized applications. Having JavaScript the only instance where this convention goes out the window does feel a bit weird.

Thirdly, it's not DRY. Each new layout will lead you to copying the same partial or code snippet with echoing variables.

Now, compare this to simply adding another line to `.env` file. Don't you get how much more natural that is?

QWp6t:
> @QWp6t You're assuming most users of Mix are committing their compiled assets to version control (along with these injected variables) which is not exactly the case, despite this being the default in Laravel installation. In fact, even laravel.com doesn't follow this convention. A lot of front-end engineers will frown on this and having worked with several Laravel projects, I've seen most people explicitly gone out of their way to ensure compiled stuff was out of version control. This isn't even the issue that should be addressed by abusing VSC. It's the responsibility of the CI and your devops if you don't want to build your scripts in production.

I made no such assumption. By most accounts, it's terrible practice to commit compiled assets or code into VCS, and I would never suggest or assume that someone is doing that. Just wanted to clarify that real quick.

-----

I'm not picking on JavaScript, really. The language is irrelevant. This has everything to do with where the variable needs to be defined. There's just not very many scenarios in which a variable used in client-side code would need to be defined at runtime as opposed to compile-time or within the source code itself.

On that note, "compare this to simply adding another line to `.env` file" could be replaced with "add another line to your build script." Unless you're committing your `.env` file to VCS, the latter makes more sense to ensure consistent builds regardless of environment.

Like I said earlier, I'm not really opposed to this PR since it doesn't affect me. I'm simply opposed to some of the reasons you guys are giving for its use.

adiachenko:
> could be replaced with "add another line to your build script." Unless you're committing your .env file to VCS, the latter makes more sense to ensure consistent builds regardless of environment.

Can you clarify what you mean by build script here? Do you mean tools like Jenkins?

And how that would apply to your workflow. Suppose your have variable X. How would you approach using it on both development and production assuming you have variable value between machines and environments.

Sorry, I am a bit dense here. I am genuinely trying to understand how the suggested alternative works and whether it's really better.

aaronhuisinga:
This would definitely be handy in some cases. We're using Bugsnag, and in order to set our API key and environment (so our development exceptions don't fill our Bugsnag logs) it would be great to use our .env file, since our compilation actually happens on the production servers versus being committed to source.

Right now we have to customize the actual NPM compilation scripts on the servers, and that could be solved by using this instead. Much simpler, IMO.

ovanschie:
I also like to see this. For Elixir we created a plugin (https://github.com/appstract/laravel-elixir-env) but build-in support from mix would be nice :)

JeffreyWay:
Okay this is merged into master now. Thanks, @adiachenko!

If somebody wants to test it out and submit a docs PR to this repo, that'd be great. :)

laugharn:
@adiachenko would you be opposed to a PR that moved the regex to Options.js?

adiachenko:
@laugharn I don't really care, you can PR it if you want, but if you're wondering about my opinion I think that the key prefix should be the same for everyone to avoid needless confusion. It's not something I would make configurable.

ruchern:
I love this PR. It will come in handy one day to some.

mirzap:
@JeffreyWay Can you tag a new release... @adiachenko  Thanks for the PR.

JeffreyWay:
New release will be tagged on Friday.

pcfutures:
This looks great, but how do we actually use it? Can't find any documentation...

I've got a file `config.js` where I'd like to do the javascript equivalent of `env('APP_URL')` - how do I do this?

adiachenko:
@pcfutures I added a PR to [5.4 docs](https://github.com/laravel/docs/pull/3288). You can read it up there for now.

pcfutures:
Doesn't appear to be working:
```JavaScript
api: 'https://api.' + process.env.APP_URL,
```
compiles to:
```JavaScript
api: 'https://api.' + __webpack_require__.i({"NODE_ENV":"development"}).APP_URL,
```
not like
```JavaScript
api: 'https://api.example.com',
```
as expected - assuming `.env` -> `APP_URL` = `example.com`

adiachenko:
Read the notes carefully. Besides `NODE_ENV` only prefixed variables are substituted (meaning you should create `MIX_APP_URL` and access it via `process.env.MIX_APP_URL`). 

Prefix is enforced so that even `process.env` can be accessed without security and memory concerns (but you should directly reference the variable anyway). This mirrors what facebook has done with their CRA repo.

pcfutures:
Ah, okay. Makes sense.

Only problem now is I have to run `npm run production` on server to compile scripts with the right `env` variables.

adiachenko:
> I have to run npm run production on server to compile scripts with the right env variables.

@pcfutures Yes, this is correct.

> which kind of defeats the point of having a .env

Not really. If you have robust CI pipeline that compiles assets in-between development and production environments, sure, you'd probably want to use something that actually accounts for environment variables at, so to speak, designation. 

But otherwise you'd probably need to build assets in production anyway, so I don't see an issue here.

pcfutures:
@adiachenko My problem with that is that my live environment is a shared hosting cpanel site that doesn't have access to a console, let alone installing npm etc.

So my only option is to switch to the live `.env` in my dev environment, run `npm run production`, then switch back to the dev `.env` - unless you have a better suggestion?

adiachenko:
> So my only option is to switch to the live .env in my dev environment, run npm run production, then switch back to the dev .env - unless you have a better suggestion?

@pcfutures Yes, if you commit your assets to version control I recommend using [Git Flow](https://support.gitkraken.com/repositories/git-flow) rules and only build production assets on master branch while tagging the particular release that will go live.

pcfutures:
@adiachenko My server doesn't connect to Git control - I move it over manually, so that's not an option.

I guess the `.env` switching way will do.

Thanks for the help.

PR #469, opened 2017-02-25
Support for vendor asset subfolders (Fixes #286)

### LINKED ISSUE #286: Folder structure for images
In the latest version images are sent to the public folder under the `images/` directory. This is great and an improvement on being placed into the root of the public directory.

However I'm wondering if there is configuration options to allow the path to be copied, for example if I have this structure:

- images/
  - vendor/
    - lightbox/
      - close.png
    - image2.png
  - image1.jpg

These are currently all sent to the `images/` folder.  I'd like the ability to copy the structure over too, apparently the `[path]` or `[folder]` can be used but this results in:

Using `[path]`
images/resources/assets/images/vendor/lightbox/close.png

Using `[folder]`
images/lightbox/close.png

Is there anyway of getting only a substring of the path (so it excludes `resources/assets/images` in the `[path]`)?

I realise I could simply move my images folder to the project root... but it would be nice to resolve this without doing that.

Thanks (I hope this is the best place to post this, otherwise I could move it to a forum..)

KristofMorva:
Currently all assets are just dumped into `fonts` and `images`. However, multiple vendor packages can use different `loading.gif` files for example, which are being overridden by each other, so as a result only one of the packages will work as intended (fixes #286).

To solve this problem, first I have updated Webpack's loader-utils to give us the ability to change the result filename any way we want (https://github.com/webpack/loader-utils/pull/59), and with its help in this PR I propose generating a relatively logical folder structure for vendor assets starting from the name of the package, ignoring image, font and asset folders. For example:

- node_modules/test/images/loading.gif => images/vendor/test/loading.gif
- bower_components/awesome/assets/font/glyph/glyph.otf => fonts/vendor/awesome/glyph/glyph.otf

### DIFF
--- a/setup/webpack.config.js
+++ b/setup/webpack.config.js
@@ -143,7 +143,7 @@ module.exports.module = {
             options: {
                 name: path => {
                     if (!/node_modules|bower_components/.test(path)) {
-                        return 'fonts/[name].[ext]?[hash]';
+                        return 'images/[name].[ext]?[hash]';
                     }
 
                     return 'images/vendor/' + path.replace(/((.*(node_modules|bower_components))|images|image|img|assets)(\/|\\)/g, '') + '?[hash]';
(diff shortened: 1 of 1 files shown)


JeffreyWay:
Looks good to me! Nice work.

PR #471, opened 2017-02-25
Fix minify when processCssUrls option is false

Rigorbb:
I have the problem with minifying css if processCssUrls option is false.

Minify is not working because raw-loader (which use in this case) does not support minify. Can we use css-loader in this case too?

### DIFF
--- a/src/Preprocessors/Preprocessor.js
+++ b/src/Preprocessors/Preprocessor.js
@@ -68,7 +68,7 @@ class Preprocessor {
         let sourceMap = Mix.sourcemaps ? '?sourceMap' : '';
 
         return [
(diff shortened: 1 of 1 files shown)


JeffreyWay:
Your PR would break things, so I've pushed a separate commit to fix this. Thanks!

Rigorbb:
@JeffreyWay are you shure that the url option may be turn off with sourceMap option enabled? Look this issue https://github.com/webpack-contrib/css-loader/issues/44

JeffreyWay:
I'm not sure, but everything seems to be working for me. 

If you find anything that breaks, just let me know.

PR #578, opened 2017-03-16
Fix in publicPath output

nachodd:
When compiling with webpack, this was outputing to the root directory (_eg. localhost/_).
If you have a custom directory (_eg. localhost/path/to/my/project_), then it won't work. 
So, it's better to output to the current working directory.

### DIFF
--- a/src/Mix.js
+++ b/src/Mix.js
@@ -80,7 +80,7 @@ class Mix {
             path: this.options.hmr ? '/' : this.options.publicPath,
             filename: filename,
             chunkFilename: chunkFilename,
-            publicPath: this.options.hmr ? 'http://localhost:8080' : ''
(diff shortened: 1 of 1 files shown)


JeffreyWay:
We used to have that, but I think we changed it to how it is now. Trying to remember why.

In the meantime, you can use mix.setPublicPath() to override it.

nachodd:
Tried using that, but didn't work for me. Maybe it was a cache-related
thing. Really, don't know.
Thanks anyway! Awesome work.

adiachenko:
Please, don't merge this PR.

[Async chunks loading](https://webpack.js.org/guides/code-splitting/#on-demand-code-splitting) will break under suggested `output.publicPath`.

JeffreyWay:
ok.

PR #592, opened 2017-03-20
cross-env 3.2 breaks folder structure

slavicd:
because cross-env 3.2 no longer has a "bin" directory in root, npm run dev fails in Laravel 5.4

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "clean-css": "^4.0.7",
     "concatenate": "0.0.2",
     "copy-webpack-plugin": "^4.0.1",
-    "cross-env": "^3.1.3",
+    "cross-env": "3.1.3",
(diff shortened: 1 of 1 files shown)


JeffreyWay:
This has already been addressed. You just have to update your package.json script to remove the bin directory. 

https://github.com/laravel/laravel/blob/master/package.json#L4-L8

And manually do `npm install cross-env -D`.

slavicd:
Addressed in what way? Is the new folder structure accounted for in master?
How is a developer installing Laravel 5.4 supposed to know this?

JeffreyWay:
It's already reflected in Laravel 5.4. You don't have to know anything.  

https://github.com/laravel/laravel/blob/master/package.json

PR #610, opened 2017-03-23
Simple tasks implementation

evsign:
How to use:

```js
mix.task('login-page', () => {
    mix.js('resources/assets/js/pages/auth/login.js', 'public/assets/js/pages')
        .sass('resources/assets/sass/pages/auth/login.scss', 'public/assets/css/pages')
});

mix.task('register-page', () => {
    mix.js('resources/assets/js/pages/auth/register.js', 'public/assets/js/pages')
        .sass('resources/assets/sass/pages/auth/register.scss', 'public/assets/css/pages')
});

mix.extract(['vue']);
mix.sass('resources/assets/sass/main/app.scss', 'public/assets/css');

// Combined task. Running this task will cause in running all tasks specified in second argument.
mix.task('auth', ['login-page', 'register-page']);
```
To run specific tasks: `node run dev -- --env.tasks=login-page` or `node run dev -- --env.tasks=login-page,register-page`. Or you can use special task keyword "all" for run all defined tasks.

Unfortunately, i didn't find the more convenient way for providing tasks into script. In my opinion `--env.tasks =` looks some ugly. If somebody knows how the same can be achieved in another way, please say)

### DIFF
--- a/setup/webpack.config.js
+++ b/setup/webpack.config.js
@@ -4,430 +4,437 @@ let webpack = require('webpack');
 let Mix = require('laravel-mix').config;
 let webpackPlugins = require('laravel-mix').plugins;
 
-/*
- |--------------------------------------------------------------------------
- | Mix Initialization
- |--------------------------------------------------------------------------
- |
- | As our first step, we'll require the project's Laravel Mix file
- | and record the user's requested compilation and build steps.
- | Once those steps have been recorded, we may get to work.
- |
- */
+module.exports = (env = {}) => {
 
-Mix.initialize();
+    let config = {};
 
+    /*
+     |--------------------------------------------------------------------------
+     | Mix Initialization
+     |--------------------------------------------------------------------------
+     |
+     | As our first step, we'll require the project's Laravel Mix file
+     | and record the user's requested compilation and build steps.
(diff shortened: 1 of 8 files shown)


QWp6t:
webpack freaks out if you pass unknown arguments to it, so `--env` is pretty much the only option that could be done from the command-line in mix's current form. (Technically, I believe you could use `cross-env` and `process.env` if you wanted, i.e., `cross-env tasks=login-page npm run dev`)

JeffreyWay:
Would many of you use this feature? I'm trying to think when I would personally want this.

evsign:
My humble opinion in support)

This is just possibility to build a part of your application. For example, i change only one line in styles and want to deploy this to production. With tasks i can just run specific task, which suited for buildings styles, without need to rebuild all assets (why i need to rebuild js if fix styles :thinking: ). Also i think this is more component way approach)

terion-name:
@JeffreyWay YES! Build main app and admin panel separately, for example. Or different parts of an app, if it is modular (happens sometimes), or compile separate themes

Zakay:
+1

JeffreyWay:
My instinct for now is to say no.  But we can always come back to this later.

PR #630, opened 2017-03-29
Change Mix.output() method.

tianyong90:
I think the mix.setResourceRoot() method should change the output.publicPath option value due to the method comments, but it DOES NOT. So I made a small change :smile:


JeffreyWay:
We actually don't need this, but thanks!

Issue #525, opened 2017-03-05
Can't import SASS when node_modules is included in webpack.config.js


adamszigeti:
- Laravel Mix Version: 0.8.8
- Node Version (`node -v`): 6.9.4
- NPM Version (`npm -v`): 4.2.0
- OS: MacOS Sierra 10.12.3

### Description:

When trying to import a sass file in JavaScript with style-loader (``import "./custom.scss";``), I get the following error, as if SASS loaders would not be in place:
```
 error  in ./src/custom.scss

Module parse failed: /{project_path}/mixbug/src/custom.scss Unexpected token (3:0)
You may need an appropriate loader to handle this file type.
| $bg: red;
|
| .container {
|   background-color: $bg;
| }

 @ ./src/app.js 2:0-23
 @ multi ./src/app.js ./src/app.scss
```

But I know it is working, because my Sass dependencies (mostly bootstrap and font awesome) are properly compiled. I noticed when I comment out [this line](https://github.com/JeffreyWay/laravel-mix/blob/master/setup/webpack.config.js#L131) from ``webpack.config.js``, it properly compiles (that is, if I also comment out [this line](https://github.com/JeffreyWay/laravel-mix/blob/master/setup/webpack.mix.js#L15) from ``webpack.mix.js``). But of course I do want both of them compiled, so when I toggle ``.sass()`` call in my ``webpack.mix.js`` back, i get the following error:
```
 error  in ./src/app.scss

Module build failed:
@import "~bootstrap-sass/assets/stylesheets/bootstrap";
^
      Invalid CSS after "...load the styles": expected 1 selector or at-rule, was "var content = requi"
      in /{project_path}/mixbug/src/app.scss (line 1, column 1)

 @ ./src/app.scss 4:14-389
 @ multi ./src/app.js ./src/app.scss
```

### Steps To Reproduce:

1. Initialize NPM ``npm init -y``
2. Install Mix, Vue, and Bootstrap-sass ``npm install --save-dev laravel-mix vue bootstrap-sass``
3. Copy settings from Mix: ``cp node_modules/laravel-mix/setup/** ./``
4. Setup some files: ``mkdir src && touch src/app.js && touch src/custom.scss && touch src/app.scss && touch index.html``
5. Fill files with content:

#### index.html
```html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <title></title>
</head>
<body>
  <div id="app">
    <custom></custom>
  </div>
[...]

PR #637, opened 2017-03-29
Lets you import scss files from javascript and at the same time compile your sass with the sass() call

tiran133:
Let's allow you to import .scss files into javascript files when you use mix.js() or mix.react() together with mix.sass().

See issue #525


JeffreyWay:
Thanks! :)

Issue #550, opened 2017-03-10
Users rule should take precedence over standard ones 


rderimay:
When passing a webpackConfig to Mix with loaders, the standard mix webpack config will take precedence over the user's passed ones.

Passing this
```
mix.webpackConfig({ 
module: { rules: [ { 
    test: /\.svg$/,
    loaders: ['svg-sprite-loader?' + JSON.stringify({
        name: '[name]_[hash]',
        prefixize: true
    })]
} ] } });
```
will lead to this rule to FOLLOW the standard mix config rules. As a consequence, the "/\.svg" test is useless as .svg is already catched by the standard .svg font rule of mix.

I think we should have the user's rules take precedence over the standard ones.

This will also solve #350

QWp6t:
jway, an ezpz way of doing this is with `webpack-merge`'s [`merge.smart()`](https://github.com/survivejs/webpack-merge#smart-merging) function.

JeffreyWay:
@QWp6t Thanks! Added.

rderimay:
@QWp6t and @JeffreyWay. This still does not solve the precedence of the standard mix webpack config of the user's passed one.

1. webpack-merge.smart will not merge intelligently a loader rule with `test: /\.(woff2?|ttf|eot|svg|otf)$/,` and another rule with `test: /\.svg$/,`by removing svg from the test string of the first one.
2. `module.exports = require('webpack-merge').smart(module.exports, Mix.webpackConfig));`will still make the Mix config have precedence over user's webpackConfig

Maybe there is another way to solve this problem, but I dont see any right now.

IlyaSemenov:
@rderimay I ended up creating `webpack.config.js` of my own where I patched up what laravel-mix emitted:

```js
const config = require('laravel-mix/setup/webpack.config');

// Exclude SVG from laravel-mix webpack config, we will handle them with vue-svg-loader
// See also: https://github.com/JeffreyWay/laravel-mix/issues/350
for (const rule of config.module.rules) {
    if (rule.test.toString() == '/\\.(woff2?|ttf|eot|svg|otf)$/') {
        rule.test = /\.(woff2?|ttf|eot|otf)$/;
    }
}

module.exports = config;
```

Don't forget to update your `package.json` (get rid of `--config=node_modules/laravel-mix/setup/webpack.config.js`).

The same can probably be implemented in `webpack.mix.js` alone by monkey-patching `mix.config.finalize`.

stephan-v:
[...]

Issue #350, opened 2017-02-08
SVGs Not Loading / vue-svg-loader


nathanaelphilip:
- Laravel Mix Version: 0.7.2
- Node Version (`node -v`): 7.2.0
- NPM Version (`npm -v`): 4.1.1
- OS: OS Sierra

### Description:
SVGs are getting dumped into the fonts folder. I’m using the `vue-svg-loader` package, and have this rule applied:

```
{
test: /\.svg$/,
loader: 'vue-svg-loader'
},
```

in my modules I’m using this: `import IconCalendar from '../../../svg/menu circle grid.svg';`

error: 

```
[Vue warn]: Invalid Component definition: /fonts/menu circle grid.svg?3644f1f735844b4b08228cc5b11a4534 
(found in component <hey-select> at /Users/nathanaelmoody/Sites/heystack-app/assets/js/components/inputs/Select.vue)
```

### Steps To Reproduce:

1. install this: https://github.com/visualfanatic/vue-svg-loader
2. add the rule
3. try to reference an svg file

Side Question: is there a way to reference everything from the project root? So I don’t have to keep doing `../../../`. 

Thanks!

nathanaelphilip:
removing *svg* from the font test “fixes” the issue. is there a way to exclude a specific path in that rule?

rderimay:
Same problem here when wanting to add this tho the webpackConfig.rules:
```
{
	test: /\.svg$/,
	loaders: ['svg-sprite-loader?' + JSON.stringify({
		name: '[name]_[hash]',
		prefixize: true
		})]
}
```

The default |svg| loader in mix will take precedence.

JeffreyWay:
Yeah, hmm... I need to figure out what we should do here. So...

1. We want SVGs to go to an /images folder.
2. But if the SVG is a font, it should go to the /fonts folder.

I'm not sure how we do that. Ideas?

JeffreyWay:
One option, I suppose, is to just defer to the folder name of the resource.

```js
    {
            test: /\.(png|jpg|gif|woff2?|ttf|eot|svg|otf)$/,
            loader: 'file-loader',
            options: {
                name: '[folder]/[name].[ext]?[hash]',
                publicPath: Mix.resourceRoot
            }
        }
```

rderimay:
Yes that would be fine, but we also want to be able to overwrite the standard loader/configs for a particular type of file.
For example, I would like to use this for icon SVGs (not for font SVGs of course...):
[...]

PR #672, opened 2017-04-06
Removed SVG from webpack test

stephan-v:
I feel that putting SVG files in a fonts directory by default is an assumption that should not be made as a default setting. 

Such a file extension does not define a font and should therefore be implemented by a user themselves.

### DIFF
--- a/setup/webpack.config.js
+++ b/setup/webpack.config.js
@@ -164,7 +164,7 @@ let rules = [
     },
 
     {
-        test: /\.(woff2?|ttf|eot|svg|otf)$/,
+        test: /\.(woff2?|ttf|eot|otf)$/,
         loader: 'file-loader',
         options: {
             name: path => {


stephan-v:
If people are downvoting please provide some kind of comment to explain why. I would love to know the reason behind it, especially since users are forced to overwrite the Webpack configuration with a partial Webpack config file of their own.

Simply using `mix.webpackConfig()` will not suffice to deal with this. Users who simply want to incorperate SVG images into their project are left to hang.

adiachenko:
I don't want to bother with redefining `file-loader` rule just because someone thinks that SVG files don't belong to the `fonts` directory. It doesn't even matter where the compiled files go. Everything might as well go the the single `dist` directory while hashing the names to avoid collisions. Please, don't overthink this.

stephan-v:
This is not about your personal preferences though. This is about whether the majority of developers who use this, implement SVG as fonts or as images.

I think the latter applies here.

adiachenko:
@stephan-v Well, my point is it's a compiled path, it isn't meant for human interaction anyway. Why does it matter to people who use SVG as images how webpack stores them internally?

It will be easier to agree with you if you describe specific problem that you encounter and then we go, like, *yeah, that makes sense*. 😄

adiachenko:
Honestly, it doesn't even matter that much, because I wager a lot of people who use SVG in their Vue components or whatever simply place the code inline rather than reaching to file-loader. I that sense, I guess, I can go along with this PR. The problem is that for some existing code bases the upgrade might be a minor annoyance and we don't seem to be solving any real issue here.

stephan-v:
There are more people who ran into this as an issue though:

https://github.com/JeffreyWay/laravel-mix/issues/350
https://github.com/JeffreyWay/laravel-mix/issues/550

adiachenko:
@stephan-v Aha, now I get it! You might want to place more emphasis on these in your first post. Specifically, a situation where the user defines their own rules for SVG files and the whole thing boggles down in confusion due to existing rule.

stephan-v:
I see there are still pull requests open from jan 4th by the way. I do not want to be a nuisance here, but it would be nice if there were either rejected(with a valid reason) or merged.

JeffreyWay:
This would break everything though. So until we have a better alternative, I'm going to close this.

stephan-v:
@JeffreyWay Care to go a little more in depth here? I am wondering what exactly will break. Maybe I can provide a better solution for you. I will dive more into the codebase tomorrow.

stigdev:
I totally agree with you @stephan-v, if there is an .svg file in my resources directory somewhere that does not mean it's a font. i actually excluded resources directory like this 
```javascript
if (! /node_modules|bower_components|resources/.test(path)) {
                    return 'fonts/[name].[ext]?[hash]';
                } 

                return 'fonts/vendor/' + path
                    .replace(/\\/g, '/')
                    .replace(
                        /((.*(node_modules|bower_components|resources))|fonts|font|assets)\//g, ''
                    ) + '?[hash]';
```
so it was node_modules|bower_components and i just added resources alongside, it is an ok "fix", it does not cause errors while compiling.

stephan-v:
I still can not see what exactly this is going to break. I feel this might break your personal implementations, but this goes beyond the needs of the few and is about how the majority of users are going to use this. 

Even if this is not the it is currently impossible to override this without creating a webpack config file yourself.

stephan-v:
@JeffreyWay Since you said "everything would break" could you please provide a workflow where this is the case so we can all find a solution here.

tpraxl:
+1 for changing this. (It's the second time, I have to deal with this issue in just a few weeks).

The vast majority of SVG files are images and not fonts. So, in case someone needs to deal with svg fonts, these fonts would end up in the image folder or would require an additional rule, which isn't perfect, but the vast majority of files would end up in the right directory.

Alternative Proposal (just a quick idea, I hadn't have the time to verify that this approach would actually work):

Our problem is having different processing requirements for the same filetype. I would suggest to define a convention for telling svg-fonts apart from svg-images: 

Something like this for fonts: 

```
test: /font.*\.svg$/
```

So the font file-loader rule in webpack.config.js would look like this:

```
test: /\.(png|jpg|gif|woff2?|ttf|eot|font.*\.svg|otf)$/
```

The (supposed) effect:

svg fonts must either be stored in a font directory (not sure if the folder name is taken into account by webpack rule tests) or carry "font" in their filename. If you need to deal with fonts from node_module packages, you might be lucky, because the expression already matches a common convention or you would need to `mix.copy('~the-node-lib/f-o-n-t/f-o-n-t.svg', 'resources/assets/fonts')` and refer to the local copy of that font.

If there are special requirements for images (like sprite-loader, svg-minification, etc), the individual custom rules could very well define something like this:

```
test: /\.sprite.\.svg$/ 
```

or 

```
test: /im[a]?g.*\.svg$/
```

Hopefully, I could make myself clear. I would very much appreciate if you could consider my proposals and I'm willing to help as much as my time allows.

stephan-v:
@tpraxl Sounds like solid proposals but I am afraid nothing will change here since I have been asking for examples since April. I am starting to get the feeling that the only workflow that will actually "break" like described above by mr. Way is his own workflow.

I seriously can not wrap my mind around why the default for SVG's are set up like this.

tpraxl:
@stephan-v Well, it certainly would break existing projects that deal with svg fonts or that already have workarounds for this issue, so this should be a matter of the next major release as I see it. (Though I would love to have it changed as soon as possible). 

@JeffreyWay: I would very much appreciate if something like this would make it into the next major release.

For anyone who comes across this thread and tries to solve this issue – it works for me this way:

Copy `node_modules/laravel-mix/setup/webpack.config.js` to your project root.

Change two rules:

```js
-        test: /\.(png|jpe?g|gif)$/,
+        // override standard mix configuration to include svgs from an img or images directory
+        test: /(\.(png|jpe?g|gif)$|im[a]?g.*\.svg$)/,
[…]
-        test: /\.(woff2?|ttf|eot|svg|otf)$/,
+        // overrides standard mix configuration that whould have included all svgs.
+        // We just test for svg that either is in a font folder or carries font in its name and exclude the others.
+        // The default configuration causes problems: 1. svgs are stored in the wrong folder. 2. When adding svg-rules for images,
+        // references refer to the fonts output directory, not to the images output directory.
+        test: /(\.(woff2?|ttf|eot|otf)$|font.*\.svg$)/,
```

That's it. Now we can tell images and fonts apart by their folder locations or names. You just need to make sure to follow this convention. So drop your assets in the respective folders or provide appropriate names.

mattkoch614:
@tpraxl - Where did you add these rule changes in ``webpack.config.js`` after you copied it into the project root?

tpraxl:
@mattkoch614 this comment is about an older version mix. Most likely, the config files have changed since then. I added "-" and "+" to my snippet to show which lines have been replaced. "-" denotes the original line, "+" the change. Sorry, but I'm currently very busy and won't be able to help you soon with more recent versions.

mattkoch614:
@tpraxl - No worries. I ended up going with a different solution:

https://github.com/JeffreyWay/laravel-mix/issues/550#issuecomment-332824496

PR #673, opened 2017-04-06
Cleared up index dependencies

stephan-v:
Although instantiating the classes directly with the require might save some line space it does not provide any overview at all.

This is especially worrying when you come across a class like `Dispatcher` and try to look for `new Dispatcher()` within the project scope. You won't find anything since the syntax is: 

```JS
new (require('./Dispatcher'));
```

Instead of:

```JS
new Dispatcher();
```

Ofcourse you can hunt the code for the `require` but it might be in multiple places for some classes.
Preferably the plugins might be split off from this file later on to clean it up a bit further.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,15 +1,24 @@
-global.options = require('./Options');
-global.entry = require('./Entry');
-global.path = require('path');
-global.Paths = new (require('./Paths'));
-global.events = new (require('./Dispatcher'));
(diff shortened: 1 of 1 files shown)


JeffreyWay:
I don't mind either version honestly, but I think it's fine as it is for now. No real need to change it. Thanks for the PR!

PR #684, opened 2017-04-09
Added babelify options, options will be passed to env preset

szachara:


JeffreyWay:
My instinct is to hold off on this for now. 

These things are really hard, as you can imagine, because PRs come in to make every possible option configurable...and it complicates the API. I'd rather stick with a base set of recommended options.

szachara:
And what about to pass it as third option to mix.babel() and configurable will be only modules option, something like:
```
mix.babel([
    'public/js/admin.js',
    'public/js/dashboard.js'
], 'public/js/all.js', {
    'modules': false
});
```

Issue #688, opened 2017-04-11
vue-loader postLoaders option with webpackConfig()


rhyek:
I have a use case where I need to post process my .vue components' compiled templates with Babel. I am targetting Chrome 49 (last Windows XP version) and UglifyJS (production). I'm using things such as `Object.values` and these generate runtime errors since they don't exist. Post processing the compiled render function fixes this.

My `.babelrc`:
```
{
  "presets":[
    ["env", {
      "targets": {
        "chrome": 49,
        "uglify": true
      },
      "useBuiltIns": true
    }]
  ],
  "plugins": [
    ["transform-runtime", {
      "polyfill": true,
      "regenerator": true
    }],
    "transform-object-rest-spread"
  ]
}
```
I currently have to create a `webpack.config.js` file in my project's root and add the following to the `vue` rule:
```js
        postLoaders: {
          html: 'babel-loader'
        }
```
I tried adding that using `webpackConfig` and doing:
```js
mix
  .webpackConfig({
    module: {
      rules: [
        {
          test: /\.vue$/,
          loader: 'vue-loader',
          options: {
            postLoaders: {
              html: 'babel-loader'
            }
          }
        }
      ]
    }
  })
```
But that obviously doesn't work since it will just appended to the `rules` array and override the original `vue` rule.

What would be the recommended way to do this without having a modified copy of `webpack.config.js` in my root?

rhyek:
Thanks, man.

PR #689, opened 2017-04-11
finalizeWebpackConfig

rhyek:
Fixes https://github.com/JeffreyWay/laravel-mix/issues/688.

This allows me to do the following in my mix config file:
```js
mix
  .finalizeWebpackConfig(config => {
    const rule = config.module.rules.find(rule => rule.loader === 'vue-loader')
    rule.options.postLoaders = { html: 'babel-loader' }
  })
```

### DIFF
--- a/setup/webpack.config.js
+++ b/setup/webpack.config.js
@@ -417,12 +417,12 @@ module.exports.plugins = plugins;
 
 /*
  |--------------------------------------------------------------------------
- | Mix Finalizing
+ | Mix Finalizing - Part 1
  |--------------------------------------------------------------------------
  |
- | Now that we've declared the entirety of our Webpack configuration, the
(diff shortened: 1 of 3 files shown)


JeffreyWay:
I'd rather not extend the API even further for an isolated issue like that. Going to close this for now, unless we have a really good reason to merge it.

rhyek:
@JeffreyWay, what other method do you propose for a use case like the one described in #688? I think this pull request provides a way to handle that and any other very specific situations where the only solution is either having your own webpack.config.js or providing some very esoteric API method (which I don't believe this is).

rhyek:
@JeffreyWay I think if it is your desire to not extend the API any further, this will provide the bare metal access needed by anyone in the future where their very specific requirements can't be handled any other way, but would prefer not to maintain their own Webpack config file.

Issue #223, opened 2017-01-29
Sourcemaps off/on


FDiskas:
Source-maps should be available only in development. On productions it should be removed.
So I suggest respect the ENV

```js
mix
  .js('resources/assets/js/app.js', 'public/js')
  .sourceMaps(!mix.inProduction)
````

JeffreyWay:
Thanks! It'll disable them automatically behind the scenes now.

growlead:
mix.inProduction - FUNCTION

mix
  .js('resources/assets/js/app.js', 'public/js')
  .sourceMaps(!**mix.inProduction()**)

PR #704, opened 2017-04-14
Allow sourcemaps to be generated for production

bbluemel:
Adding always parameter to generate separate file sourcemaps when in production mode.

Coming from elixir which had the `elixir.config.sourcemaps = true;` option to always generate sourcemaps even in production was useful for us, since we use Sentry for providing insights into errors in JS Code. :)

### DIFF
--- a/src/Api.js
+++ b/src/Api.js
@@ -274,9 +274,11 @@ class Api {
 
     /**
      * Enable sourcemap support.
+     *
+     * @param {boolean} always
      */
-    sourceMaps() {
-        global.options.sourcemaps = (this.Mix.inProduction ? false : '#inline-source-map');
+    sourceMaps(always = false) {
+        global.options.sourcemaps = ( this.Mix.inProduction ? ( always ? '#source-map' : false ) : '#inline-source-map' );
 
         return this;
     };


ankurk91:
This will be a nice addition. I am using bugsnag js in my project and bugsnag [recommend]( https://docs.bugsnag.com/platforms/browsers/source-maps/) have source map available even in production.

It would be good if we can generate **hidden-source-map** source maps.
This is the [recommended](https://webpack.js.org/configuration/devtool/) by webpack.

bbluemel:
It looks like you can use hidden source maps with Sentry as well, so seems like a good option, but not sure making it the default is the "right" decision.

I wonder whether a better option would be to have something like
```
mix.options({
    productionSourceMaps: true // for standard inclusion, 'hidden' for hidden, or allowing the full text, e.g. 'cheap-source-map'
});
```
Giving the developer the options of what to use.

rhyek:
This is a great addition, but IMO the default for `always` should be `true` since I believe source maps in general are mostly useful and pretty common in production environments. Thinking you won't need source maps for tracking down errors manually or with tools such as sentry once you ship your code is very misguided, I think.

With regards to using `hidden-source-map` optionally, you can already do this anyways using `webpackConfig`. I'm currently doing the following:

```js
const production = process.env.NODE_ENV === 'production'
const devtool = production ? 'source-map' : 'inline-source-map'

mix
  .webpackConfig({
    devtool
  })
  ...
```

I mean, the last thing you want is for `laravel-mis` to have an API as complex as `webpack` 😄.

The default for production should definitely be `source-map`.

bbluemel:
@rhyek I love your solution to the problem there, I may have to adopt that.  
I've been holding off merging my elixir to mix conversion, because I don't want to lose sourcemaps in production, and without something like the above, it certainly would be a show stopper for all the sites I use elixir on.

Considering #223 there appears to be some divide in the community on what should be the standard.  I agree of course that source-maps in production should exist by default.  However, I understand that there are some out there that use the minification as a way of protecting their code.

I would imagine the 80% use case that mix aims for is probably more in the camp for generating in production rather than not.

ankurk91:
Lets disable the sourcemaps in production and inline in dev mode by default.
Let anyone override default behaviour in this way -
```js
// Default behaviour for new users
mix.sourceMaps();

// For more advanced user
mix.sourceMaps('any-devtool-option-here');
```
one can pass any valid string from [here](https://webpack.js.org/configuration/devtool/#devtool)

rhyek:
Why should the default for production be to disable source maps when I, as a developer, am specifically calling `sourceMaps()`? IMO, a less "advanced user" wouldn't even be calling `sourceMaps()`, and those who have a very specific reason to disable them *only* for production are probably less than those who would like them available for all environments as the default behavior for the method. So, use the option for those less common scenarios, no?

rhyek:
I think in general, the expectation when using `sourceMaps()` is that they're just always generated. I actually spent a couple minutes figuring out why they were being generated locally, but not on my server. 
The option should be made available for people who are actively trying to acomodate different scenarios on their end.

bbluemel:
I agree with @rhyek that the expectation would be for them to be generated when calling the function.

And I went the route of trying to minify in development first, which of course didn't work as the source maps were still in-line creating larger than expected files.

elixir didn't have a "sourceMaps" function, it just did it on dev, and not on prod unless you specified `elixir.config.sourcemaps = true;`, so calling the function here is also akin to stipulating that you want sourceMaps.

marvinosswald:
Any progress ?

JeffreyWay:
I've merged this, though it sucks that it has to be `npm run development`. I know people will try `npm run dev` and wonder why it's not working. :/

PR #738, opened 2017-04-24
Allow different types of source mapping, by default `#inline-source-map`

chrispelzer:
This allows for multiple ways to use sourceMaps without needing to add custom webpack options.
https://webpack.github.io/docs/configuration.html#devtool

The sourceMapFilename is by default [filename].map if the source map is a separate file using `#sourcemap`


ankurk91:
Also see #704

Issue #200, opened 2017-01-27
Unexpected character '@' after upgrade from v0.5.3 to v0.5.9


rn2ee:
Upgrading from v0.5.3 to v0.5.9 causes the following errors:

Module parse failed: ...resources\assets\sass\portal.scss Unexpected character '@' (1:0)
You may need an appropriate loader to handle this file type.
| @font-face {
|     font-family: 'Open Sans';
|     font-style: normal;

Switching back to v0.5.3 or v0.5.7 without any other config changes everything is ok...

poldixd:
Hi, i got the same error. I've updated node.js and npm and deleted the `node_modules` folder and reinstalled the packages with `npm install`. Still the same error...

```
 ERROR  Failed to compile with 1 errors

 error  in ./resources/assets/sass/app.scss

Module parse failed: D:\xampp\htdocs\laravel5\resources\assets\sass\app.scss Unexpected character '@' (2:0)
You may need an appropriate loader to handle this file type.
| // Bootstrap
| @import "node_modules/bootstrap-sass/assets/stylesheets/bootstrap";

 @ multi ./resources/assets/js/app.js ./resources/assets/sass/app.scss

npm ERR! Windows_NT 10.0.14393
npm ERR! argv "C:\\Program Files\\nodejs\\node.exe" "C:\\Program Files\\nodejs\\node_modules\\npm\\bin\\npm-cli.js" "run" "dev"
npm ERR! node v6.9.4
npm ERR! npm  v4.2.0
npm ERR! code ELIFECYCLE
npm ERR! errno 2
npm ERR! @ dev: `cross-env NODE_ENV=development webpack --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js`
npm ERR! Exit status 2
npm ERR!
npm ERR! Failed at the @ dev script 'cross-env NODE_ENV=development webpack --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js'.
npm ERR! Make sure you have the latest version of node.js and npm installed.
npm ERR! If you do, this is most likely a problem with the  package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     cross-env NODE_ENV=development webpack --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs
npm ERR! Or if that isn't available, you can get their info via:
[...]

PR #749, opened 2017-04-26
Fix for Issue 200 failing to parse scss on Windows

HarrisJT:
This fix comes from https://github.com/JeffreyWay/laravel-mix/issues/200 in a post by Datazource. The thread contains many peoples issues and solutions. This one works for me and was wondering if this affected anyone else.

Change :
    test() {
        return new RegExp(this.src.path.replace(/\\/g, '\\\\') + '$');
    }
To :
    test() {
        return new RegExp(this.src.file.replace(/\\/g, '\\\\') + '$');
    }

### DIFF
--- a/src/Preprocessors/Preprocessor.js
+++ b/src/Preprocessors/Preprocessor.js
@@ -58,7 +58,7 @@ class Preprocessor {
      * Get the regular expression test for the Extract plugin.
      */
     test() {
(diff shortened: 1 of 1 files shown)


JeffreyWay:
We can't only rely on the file name, because it could potentially match multiple files.

PR #765, opened 2017-05-01
Added docs for using stats object with event hooks

jeffturcotte:
Just a quick addition to the event hook docs showcasing the ability to use/inspect Webpack's Stats object. Very useful for scripting any manual compilation/processing steps.

Thanks!


JeffreyWay:
Thanks!

PR #768, opened 2017-05-01
Add reference to 'copy-webpack-plugin' in package.json

SethTompkins:
When running `npm run test` in the repo it barks about missing `copy-webpack-plugin`. This adds it as a dep in package.json.


JeffreyWay:
I think I need to fix a couple tests related to copying, but we don't use that package anymore in the codebase.

PR #769, opened 2017-05-02
Disable ExtractTextPlugin for preprocessors when Hot Module Reloading

SethTompkins:
This allows webpack dev server to appropriately inject css updates. Currently css refreshing outside of your js bundle does not work properly with Hot Module Reloading.

### DIFF
--- a/src/Preprocessors/Preprocessor.js
+++ b/src/Preprocessors/Preprocessor.js
@@ -30,7 +30,10 @@ class Preprocessor {
      */
     getExtractPlugin() {
         if (! this.extractPlugin) {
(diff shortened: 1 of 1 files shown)


JeffreyWay:
I think this will break projects for people who expect that file to be created. 

My instinct is to just push mix.browsersync() instead, since it can be used in more places.

SethTompkins:
Would it be acceptable if I made an MR that created a separate option to disable `ExtractTextPlugin` on demand? I tested HMR with browserSync and the two do not work together for reloading css and JS. Creating one more hoop to jump through in order to disable the plugin would keep people who don't understand the consequences from being confused.

Issue #426, opened 2017-02-19
No image-min for mix?


goodbytes-gb:
not really an issue, but moving to Laravel 5.4/Mix from 5.3/Elixir seems to have killed my image min functionality as it relies on Gulp.

trying to stick with a web pack/mix work flow what are options for automatic image min?

previously I would copy images to resources/images and they would be optimised and copied to public/images.

Thanks

JeffreyWay:
I haven't decided yet if we should include image optimization with the default Mix install.

But of course, you can always manually add it in with `mix.webpackConfig()`.

https://github.com/Klathmon/imagemin-webpack-plugin

alenabdula:
I understand this issue has been closed for some time but wanted to leave this here in case anyone is trying to configure image optimization with Laravel Mix.

`npm install --save-dev imagemin-webpack-plugin copy-webpack-plugin imagemin-mozjpeg`

```js
const { mix } = require('laravel-mix');
const ImageminPlugin = require('imagemin-webpack-plugin').default;
const CopyWebpackPlugin = require('copy-webpack-plugin');
const imageminMozjpeg = require('imagemin-mozjpeg');

mix.webpackConfig({
    plugins: [
        new CopyWebpackPlugin([{
            from: 'resources/assets/images',
            to: 'img', // Laravel mix will place this in 'public/img'
        }]),
        new ImageminPlugin({
            test: /\.(jpe?g|png|gif|svg)$/i,
            plugins: [
                imageminMozjpeg({
                    quality: 80,
                })
            ]
        })
    ]
});
```
Hope that helps.
Best,
Alen

goodbytes-gb:
@alenabdula perfect thanks :)

JeffreyWay:
@alenabdula Image optimization is included in the next tagged release of Mix.

valh1996:
@JeffreyWay 

Hello, how do we put the images in version please? I would like to enjoy the cache busting also with pictures :)

My code:
`mix.styles([
    'resources/assets/css/custom.css'
  ], 'public/css/all.css');

mix.copy('resources/assets/img/*.*', 'public/img');

if (mix.config.inProduction) {
    mix.version();
}`

ralphschindler:
@JeffreyWay the above example is what I am attempting to solve as well, essentially a way to

- mix.copy() images as a glob path 
- make them version capable
- manifestable (so that the mix() function will write the proper file name during production & non-hashed in local env).

[...]

PR #781, opened 2017-05-06
add imgLoaderOptions

leo108:
we can disable the `imagemin` action to reduce build time. To enable it:

```
mix.options({
    imgLoaderOptions: {
        enabled: true,
        gifsicle: {},
        mozjpeg: {},
        optipng: {},
        svgo: {},
    }
})
```

https://github.com/JeffreyWay/laravel-mix/issues/426#issuecomment-299649992


JeffreyWay:
But now you're disabling everything by default, right? We don't want to do that.

Issue #783, opened 2017-05-07
[Feature request] Custom path for CSS URL processing


GustavsR:
- Laravel Mix Version: 0.11.4
- Node Version (`node -v`): 7.9.0
- NPM Version (`npm -v`): 4.2.0
- OS: Windows 10 Pro (Version: 1703)

### Description:
If we have relative paths in our CSS (and `processCssUrls` is `true`), then after the processing, these files are moved to `/public/images` or `/public/fonts`. Some people may not like that the files are moved to the root directory and want to move the files to the `/public/assets/images`, for example.

This path is hardcoded in `setup/webpack.config.js`:
```
if (! /node_modules|bower_components/.test(path)) {
    return 'images/[name].[ext]?[hash]';
}

return 'images/vendor/' + path
    .replace(/\\/g, '/')
    .replace(
        /((.*(node_modules|bower_components))|images|image|img|assets)\//g, ''
    ) + '?[hash]';
```

### Steps To Reproduce:

There is no support for this.

A new config variable could be added to move the files to a custom directory.

JeffreyWay:
I'd rather just stick with a common default for now.

TheKalpit:
+1 for the issue

Just ran into this situation. The website is located in a sub-directory, so it is absolutely necessary to add "/base/path/" before the generated paths.

pxwee5:
+1

stefro:
+1

andonovn:
+1

Froxz:
+1

PR #784, opened 2017-05-07
Add an option to specify custom processed URL path

GustavsR:
Issue #783


GustavsR:
This is (probably) not the best fix, but it works.
```
  .options({
    urlProcessPath: 'assets/'
  })
```

JeffreyWay:
I think we'll take a slightly different approach. But thanks for the PR!

PR #798, opened 2017-05-11
Set the sourcemap option for the sass plugin according to the current configuration

tolnaiz:
The sourcemap option shouldn't be always true for the sass plugin, instead we can set it based on the mix configuration.


JeffreyWay:
If I remember correctly, it *always* has to be true in order for the resolve url loader plugin to work properly.

PR #802, opened 2017-05-12
Small fix to setup.

lordraz0r:
Commented out a potentially confusing line that might leave developers that are new to Mix scratching their heads over the error on a fresh project it produce. It should be left in code to serve as an example but should not be uncommented.


JeffreyWay:
I think it's probably okay as it is, but thanks so much.

PR #810, opened 2017-05-16
Update "cross-env" version

arikaiy:


JeffreyWay:
I'm not sure that cross-env should be a dependency of Mix in the first place, so my hope is to remove it with the next major release.

Issue #884, opened 2017-06-16
Add support for variable expansion?


eberkund:
It would be nice to be able to do something like this in my `.env` file,

    STRIPE_KEY=xxx
    MIX_STRIPE_KEY=${STRIPE_KEY}

Laravel handles this fine because _phpdotenv_ allows [nesting variables](https://github.com/vlucas/phpdotenv#nesting-variables) but it looks like the JavaScript version used by _laravel-mix_ [does not](https://www.npmjs.com/package/dotenv#what-about-variable-expansion). This would be a very helpful addition and it looks pretty simple [to add](https://github.com/motdotla/dotenv-expand#usage).

    var dotenv = require('dotenv')
    var dotenvExpand = require('dotenv-expand')

    var myEnv = dotenv.config()
    dotenvExpand(myEnv)

JeffreyWay:
Merged.

PR #885, opened 2017-06-16
Add support for variable expansion

eberkund:
See #884


ruchern:
Why do you need to install additional packages?

Just commit the following in #884 into `.env.example`

eberkund:
This isn't for the PHP side of things, which already supports variable expansion through _phpdotenv_. 

This is for the [enviroment variables ](https://laravel.com/docs/5.4/mix#environment-variables) feature of mix which is useful for injecting constants into your JavaScript.

HSPDev:
I think this is fine. @eberkund is right, the PHP side of things doesn't matter, as the .env file is passed by JS, and I actually see this as a bug, as laravel-mix might misbehave/break, if you are already using expansion in your .env file in Laravel, and try to run Laravel mix.

ankurk91:
ping @JeffreyWay 
Can we get this merged, this gonna be very useful.

PR #894, opened 2017-06-19
#824 Added ability to disable success notifications

### LINKED ISSUE #824: [Question] How to disable notification if the build is successful?
I want to disable os notifications on success builds when I hit save. I just want be notified if there are errors or warnings.. I tried the code below but didn't work.
```javascript
mix.js('resources/assets/js/app.js', 'public/dist/vendor.js')
   .sourceMaps().then(function (stats) {
        // if (!stats.hasErrors() || !stats.hasWarnings()) {
            mix.disableNotifications();
        // }
   });
```

HSPDev:
Added ability to disable success notifications and only show errors. 
Will still show a single success notification if the previous build failed with an error, to indicate its clean now.

Fixes #824. The code was more or less built into the plugin we are using, so just had to wrap the API around it. You could possibly rename the config value if you want, but other than that I think we are clear.

### DIFF
--- a/src/Api.js
+++ b/src/Api.js
@@ -382,7 +382,15 @@ class Api {
         Config.notifications = false;
 
         return this;
-    }
+    };
+
+    /**
+     * Disable success notifications.
+     */
+    disableSuccessNotifications() {
+        Config.notificationsOnSuccess = false;
+        return this;
+    };
 
 
     /**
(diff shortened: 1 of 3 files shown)


JohnnyWalkerDesign:
Yay!!!

PR #903, opened 2017-06-20
Update references to file hashing with query string

samtsai:
Minor update to docs now that 1.0.* introduces query string versioning.


JeffreyWay:
Thanks! :)

PR #910, opened 2017-06-21
Add link to config in package.json for v1.0+

ruchern:
The reason for this is because `webpack.config.js` is no longer in the root directory by default.

Signed-off-by: Ru Chern Chong <me@ruchern.com>


ruchern:
That was real quick!

PR #927, opened 2017-06-22
Modified installation docs for installation on windows

jkniest:
I've modified the NPM scripts. The old ones do cause an error on Windows systems: "NODE_ENV not found". 

This small modification fixes it.

I wish you a nice day,
Jordan

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -59,6 +59,17 @@ As a tip, consider adding the following NPM scripts to your `package.json` file,
 
 ```js
   "scripts": {
+    "dev": "NODE_ENV=development webpack --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js",
+    "watch": "NODE_ENV=development webpack --watch --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js",
+    "hot": "NODE_ENV=development webpack-dev-server --inline --hot --config=node_modules/laravel-mix/setup/webpack.config.js",
+    "production": "NODE_ENV=production webpack --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js"
+  }
+```
+
+If you are using windows you should install the `cross-env` package via `npm install cross-env --save-dev` and change the scripts to the following. Otherwise you'll get an error that "NODE_ENV" is not found.
+
+```js
+  "scripts": {
     "dev": "cross-env NODE_ENV=development node_modules/webpack/bin/webpack.js --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js",
     "watch": "cross-env NODE_ENV=development node_modules/webpack/bin/webpack.js --watch --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js",
     "hot": "cross-env NODE_ENV=development node_modules/webpack-dev-server/bin/webpack-dev-server.js --inline --hot --config=node_modules/laravel-mix/setup/webpack.config.js",
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -59,9 +59,9 @@ As a tip, consider adding the following NPM scripts to your `package.json` file,
 
 ```js
   "scripts": {
-    "dev": "NODE_ENV=development webpack --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js",
-    "watch": "NODE_ENV=development webpack --watch --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js",
-    "hot": "NODE_ENV=development webpack-dev-server --inline --hot --config=node_modules/laravel-mix/setup/webpack.config.js",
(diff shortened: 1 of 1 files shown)


jkniest:
@ruchern Heh.. Something wrong with it? 🛩

ruchern:
@jkniest If you would like to know what I've said previously, the `scripts` in `package.json` will then be reverted to pre-v1.0 version which I am not sure if this is likely to get merged or not.

jkniest:
Ah ok. The default laravel installation does include these scripts, and the scripts of the documentation (without laravel) are not working on windows.

ruchern:
I understand the Windows part. From the way I see it, I think Windows user might have to install the `cross-env` package and set it on their own. Laravel-Mix not only being used for Laravel projects.

eberkund:
I manually made those same changes to my installation, so looks good to me. I am not sure about other OSes though, I'm on Windows.

QWp6t:
Yeah, I think out of the box the instructions should be cross-platform. Developers often collaborate on projects, and collaborators might be using different operating systems. I think it definitely shouldn't be left incompatible with Windows considering that Windows still has majority market share among operating systems.

decadence:
+1 for proper installation on Windows

devcircus:
It's a common case to have to pull in cross-env when using some packages on windows. Seems it should be left as-is. Keeping it in Mix brings along extra baggage w/maintenance concerns that Mix shouldn't be concerned with.

jkniest:
Ok, I see your point. Should I close this PR then?

jkniest:
Otherwise I agree with @QWp6t. Or maybe add another section only for windows in the docs?

ruchern:
@jkniest Yeah add a note to Windows users to say that `cross-env` is needed for it to work on Windows.

QWp6t:
@devcircus What maintenance concerns do you have?

This doesn't add any extra maintenance to laravel-mix. It simply updates the docs.

[Laravel itself comes with `cross-env` out of the box](https://github.com/laravel/laravel/blob/master/package.json#L15).

This PR parallels how things are already done in Laravel.

@jkniest The only other thing I'd add is that the docs should advise people to install cross-env. Otherwise if they copy/paste those script lines, they won't actually be able to use them.

ruchern:
@QWp6t That is assuming that you are using `laravel-mix` for Laravel projects. However, `laravel-mix` is not exclusive for Laravel projects only.

QWp6t:
@ruchern I understand that full well. The docs that are being edited in this PR are the docs for using Laravel Mix without Laravel.

ruchern:
@QWp6t Yup. So I suggest to add a separate note to mention that for Windows users, `cross-env` package is recommended or something of that sort.

devcircus:
Search for "cross-env" in the repo's issues. It's proven to be a maintenance nightmare already. It could, however become as much as a problem without it. So I see the other side. Still don't think it's mix's problem to handle as long as it's documented though.

jkniest:
I will change the docs tomorrow. I don't have much time right now.

ruchern:
@jkniest Just do not miss the Jeffrey Way's window period. He comes here at random times ;)

QWp6t:
@devcircus Again, this isn't adding any code to Laravel Mix. It's updating the docs. There are no maintenance concerns. And again, Laravel itself ships with cross-env.

I don't understand your guys' objections. The current docs don't work out of the box on the most popular operating system in the world. This PR resolves that by making it compatible with any operating system that can run node.js.

devcircus:
My most sincere apologies. I misunderstood. I think it's a great idea to cover this in the docs. I do however think that it should be made clear what problem cross-env solves. Also since many users are not using Mix in a Laravel project, it should be clear that in order to use cross-env, it needs to be installed via npm. So yes, it needs to be there but needs more detail.

jkniest:
I've added another commit which reverts the changes form the first commit and adds another small section only for windows users. Check it out

Have a nice day,
Jordan

jkniest:
@ruchern Is there anything else I need to do to get this merging? Or should I just wait until @JeffreyWay has some time to review it?

Thanks and have a nice day,
Jordan

ruchern:
Just wait for @JeffreyWay to come in.

jkniest:
Any updates on this? Ping @JeffreyWay 

Thanks and have a nice day,
Jordan :)

francescobianco:
@jkniest Nobel peace prize for!!! Why is laravel mix so called if it can be used besides laravel? @QWp6t @ruchern 
Would it be possible to fork the project and keep it under a different name?

haasz:
I have made an [extension](https://gitlab.com/haasz/laravel-mix-ext) for the Laravel Mix. I recommend its use for those developments which are stand-alone (not Laravel) projects. Its latest version (v1.4.4) already supports the Windows operating system, which solves the above problem.
The extension is available here:
https://www.npmjs.com/package/laravel-mix-ext
https://gitlab.com/haasz/laravel-mix-ext
https://github.com/haasz/laravel-mix-ext

JeffreyWay:
Sorry for the long delay, but I'd rather we just show a windows compatible example for setting env vars instead of requiring that they manually install `cross-env`.

PR #947, opened 2017-06-29
add option to prevent the manifest file from being created

ThorbenG:
My Instructor gave me the task to contribute to Open-Source projects.

While we where working on a project where we had no use for the manifest file, he thought it would be a nice addition to have the option to be able to prevent it from being created at all.

So i worked out this solution that prevents the refresh method from creating the mix-manifest.json if the option is set.

I hope i can contribute to awesome projects like this one in the future.

Best regards
Thorben

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -2,6 +2,7 @@
 
 ```js
 mix.options({
+  createManifest: true,
   extractVueStyles: false,
   processCssUrls: true,
   uglify: {},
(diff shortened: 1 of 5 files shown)


JeffreyWay:
Thanks so much, but we do need that manifest for the main compile. If folks really don't want it, they can manually delete the file after compile (within `mix.then()`).

PR #954, opened 2017-06-30
Use Vue.js ES Module build instead of commonjs

fragkp:
Personally, I think the ES-Module for Vue.js should be the recommended way.


andre-dw:
why is this reverted back to `common.js` and only uses `.esm` if i use typescript?

Issue #960, opened 2017-07-03
setPublicPath sets wrong path when using mix.js()


omnichronous:
- Laravel Mix Version: **1.0.7**
- Node Version: **6.10.2**
- NPM Version: **5.0.4**
- OS: **Windows 7 Ultimate x64**

### Description:
![image](https://user-images.githubusercontent.com/8319869/27796350-65e0a676-6012-11e7-8cf1-8fded120a563.png)

I'm trying to relocate my assets to an 'assets' subfolder inside /public. The mix-manifest.json gets created in the correct location and mix.less() outputs my CSS files and updates the reference in the manifest with the correct location as well. The mix.js() method however behaves strangely, as you can see in the screenshot.

I'm using the simplest possible setup to demonstrate this - a fresh installation of Laravel and npm install. I have tried adding and removing slashes all over the place, but to no avail.

### Steps To Reproduce:
1. Edit the webpack.mix.js file with the following:
```
mix.setPublicPath('public/assets');
mix.js('resources/assets/js/index.js', 'js');
mix.less('resources/assets/less/index.less', 'css');
```
2. `npm run dev`

JeffreyWay:
Hmm, strange. I just used your exact setup, and it created `public/assets/mix-manifest.json` that contains:

```js
{
    "/js/app.js": "/js/app.js",
    "/css/app.css": "/css/app.css"
}
```

demyan1411:
I have this issue too http://prntscr.com/fr8qqc

mix.setPublicPath(path.join(__dirname, "/dist"))
mix.js('src/assets/js/app.js', 'js')
   .extract(['jquery','bootstrap','vue','tether']);

omnichronous:
Another thing I forgot to mention is that Mix will not create the assets directory on its own - unless I create it myself I get the 'no such file' error:

`Error: ENOENT: no such file or directory, open 'D:\TRANSIT\boilerplate\public\assets\mix-manifest.json'`

---

So my buddy also tested it on a fresh install and it created the files correctly, his setup was

- Laravel Mix Version: **0.12.1**
- Node Version: **6.9.5**
- NPM Version: **4.9.5**
- OS: **Windows 7 x64**

Apparently it's something in the major version update. demyan1411's fix works for me right now but it would be nice to not edit inside node_modules.

JeffreyWay:
@omnichronous I fixed this yesterday. Just need to tag it.

omnichronous:
[...]

PR #961, opened 2017-07-04
Fix paths for windows in pathFromPublic function

demyan1411:
Hello,
1) I added path.normalize - because on windows this: ```this.filePath.startsWith(publicPath)``` not working,
2) function pathFromPublic returned '/app', and in webpack.output [name] - was '/app', so i added folderpath variable

#960


alanondra:
Looks like you still have some stray debugging code?

Issue #953, opened 2017-06-30
Wrong path after change file, while npm run watch


Andrey9300:
- Laravel Mix Version: 1.0.7
- Node Version: 8.1.0
- NPM Version: 5.0.3
- OS: Windows 10

When I run command: 
`npm run watch (watch it is: cross-env NODE_ENV=development node_modules/webpack/bin/webpack.js --watch --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js)`

I copy files from directories:
`mix.copyDirectory('./resources/assets/js/modules', './public/js/modules');`

First copy is good, but when I change files and watcher recopy files again, he change the path:
new path is join(from, to)
In file `src/tasks/CopyFilesTask.js` in method `onChange(updatedFile)` call method `append()`, which stend in file `src/File.js/` This method join paths: `return new File(path.join(this.path(), append));`

Example:
   the folder structure 'from' before build:
```
      resources/js/modules/test1
      resources/js/modules/test2
      resources/js/modules/test3
```
  the folder structure 'to' before build (empty):
```
      public/
```
  the folder structure 'to' after first build (it is ok):
```
      public/js/modules/test1
      public/js/modules/test2
      public/js/modules/test3
```
  the folder structure 'to' after change files build (it is wrong):
```
      public/js/modules/test1
      public/js/modules/test2
      public/js/modules/test3
      public/js/modules/resources/js/modules/test1
      public/js/modules/resources/js/modules/test2
      public/js/modules/resources/js/modules/test3
```
Result: path 'from' and path 'to' joined

You can try this so:
   `webpack.mix.js: mix.copyDirectory(from, to);` 
   where 'from' - hierarchical folder structure with js files 
and npm run watch
   where 'watch' - `cross-env NODE_ENV=development node_modules/webpack/bin/webpack.js --watch --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js`

Andrey9300:
problem here https://github.com/JeffreyWay/laravel-mix/blob/master/src/tasks/CopyFilesTask.js#L32
on next week I fix it

PR #962, opened 2017-07-04
path.normalize before replace in method "onChange"

Andrey9300:
It is answer on this issue https://github.com/JeffreyWay/laravel-mix/issues/953


JeffreyWay:
Thank you!

PR #996, opened 2017-07-13
Serve static files from webpackDevServer

### LINKED ISSUE #344: run hot the server root path isn't public ?
npm run hot. 

Request  `http://127.0.0.1:8080/css/app.css` got 404 error.

If I mv  `public/css/app.css` to Laravel project root path, then request  `http://127.0.0.1:8080/app.css`  is successful. 

Have any params to set the root path ?

      "hot": "node node_modules/cross-env/bin/cross-env.js NODE_ENV=development node_modules/webpack-dev-server/bin/webpack-dev-server.js --inline --hot --config=node_modules/laravel-mix/setup/webpack.config.js",

jbbr:
Adding the `contentBase` to the devServer config does fix hot reloading when using non-webpack combine & copy tasks together with hot reloading.

E.g. when using `combine()` and Laravel`s PHP `mix()` helper

This should resolve #344

Note: A side effect is, that possible PHP- and other files in the public directory are served by the dev server as well (in plain text).

### DIFF
--- a/src/builder/webpack-default.js
+++ b/src/builder/webpack-default.js
@@ -35,6 +35,7 @@ module.exports = function () {
             headers: {
                 "Access-Control-Allow-Origin": "*"
             },
+            contentBase: path.resolve(Config.publicPath),
             historyApiFallback: true,
             noInfo: true,
             compress: true,


JeffreyWay:
Thanks!

PR #1005, opened 2017-07-16
WIP: Source Maps when using combine()

jbbr:
With this PR [babel-concat](https://github.com/luigi-rosso/babel-concat) is used when using `combine()` of JS files together with `.sourceMaps()`

Inline & `.map` sourceMaps are generated for proper debugging and "jump-to-source" in JS dev tools.

I don't expect this PR to be merged in the current state but I think this feature might be interesting for others and maybe a better implementation could be merged sometime.

- babel is probably an overkill for the sole purpose of source-maps. Combining is slower.
_Combining can be disabled by setting `skipSourceMaps` to true for specific files_

- babel-concat seems abandoned and a more up-to-date fork has to be used    
_It does not contain a lot code so the sourceMap generating part could be extracted and used without babel. It uses the official mozilla [source-map](https://github.com/mozilla/source-map) lib. It could also be swapped with something like [concat-with-sourcemaps](https://github.com/floridoo/concat-with-sourcemaps)_

- babel does not work with css files.

Off Topic: Is there any documentation on how to run the tests of this repo? nyc/ava doesn't find any tests (config file not in repo?). Probably I'm doing something wrong. Some kind of contribution guide would be really helpful.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
   "license": "MIT",
   "dependencies": {
     "autoprefixer": "^7.1.1",
+    "babel-concat": "git+https://github.com/luigi-rosso/babel-concat.git",
     "babel-core": "^6.24.1",
     "babel-loader": "^7.1.1",
     "babel-preset-env": "^1.5.1",
--- a/src/Api.js
+++ b/src/Api.js
@@ -174,13 +174,14 @@ class Api {
      * @param {string|Array} src
      * @param {string}       output
      * @param {Boolean}      babel
+     * @param {Boolean}      skipSourceMaps
      */
-    combine(src, output, babel = false) {
+    combine(src, output, babel = false, skipSourceMaps = false) {
         output = new File(output || '');
 
         Verify.combine(src, output);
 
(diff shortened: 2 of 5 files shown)


jbbr:
I found a probably better way of archiving sourceMap support for legacy JS by exploiting webpack to do the concatenation, sourceMaps etc.

Just add the following webpackConfig:
```js
mix.webpackConfig({
  plugins: [
    new webpack.LoaderOptionsPlugin({
      debug: true,
    })
  ],
  rules: [
    test: /\.exec\.js$/,
    use: [ 'script-loader' ],
  ],
})
```

And install script loader: `npm install --save-dev script-loader`

Now you can simply import/require files with the ending `.exec.js` and they will be executed in the good old global scope.

The gobal loader configuration `debug: true` is required for sourceMaps to work unfortunately.
If you don't need or want sourceMaps it can be omitted.

jbbr:
I think using `script-loader` is good option for full sourceMap support.
Maybe the current `combine` could even be replaced with an implementation using webpack.

I'm closing this as the implementation is not good enough for production.

PR #1012, opened 2017-07-18
Enable standalone postcss

### LINKED ISSUE #728: Postcss support for plain css files
- Laravel Mix Version: 0.11.2
- Node Version (`node -v`): 6.10.2
- NPM Version (`npm -v`): 4.5.0
- OS: OSX El Capitan

### Description:
I have a css setup similar to [this one](https://github.com/OnsenUI/OnsenUI/tree/master/css-components/src), where a `index.css` file imports other css files that use [css variables](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_variables).

I'd need to apply [postcss-import and postcss-cssnext](http://cssnext.io/postcss/) to the `index.css` file.

### Steps To Reproduce:

There is no support for postcss on mix: https://github.com/JeffreyWay/laravel-mix/issues/617

A `mix.css(src,dist)` method could read css files, parsing the `@import` rules and applying the mix options postcss plugins.

sebastiandedeyne:
This PR adds a `postcss` method to the Mix API, which doesn't use any preprocessor except PostCSS.

Overview:
- Added `postcss` to the public API
- Files processed this way are excluded by the `css-loader` to avoid duplicate loading (which throws an error)
- PostCSS is already used for the preprocessors, so added an exception when registering the preprocessor loader.

As far as I know, this is non-breaking because it's all behind the new API method.

This PR would resolve #728.

### DIFF
--- a/src/Api.js
+++ b/src/Api.js
@@ -142,6 +142,16 @@ class Api {
 
 
     /**
+     * Register postcss compilation.
+     *
+     * @param {string} src
+     * @param {string} output
+     */
+    postcss(src, output) {
+        return this.preprocess('postcss', src, output);
+    };
+
+    /**
      * Register a generic CSS preprocessor.
      *
      * @param {string} type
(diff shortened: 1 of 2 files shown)


JeffreyWay:
Thanks, Sebastian!

PR #1059, opened 2017-07-27
Fix HMR on Windows when extracting vendor libraries

matical:
When extracting vendor libraries (following the current docs, where `extract()` is just specifying an array for the libs), vendor.js will be seemingly left out -- 404's from the node server on 8080.

Even though Windows should be able to handle different slashes, vendor.js would act as if it doesn't exist.

### DIFF
--- a/src/builder/Entry.js
+++ b/src/builder/Entry.js
@@ -63,9 +63,17 @@ class Entry {
             );
         }
 
+        let getPathJoin = (vendorPath) => {
+            if (process.platform === 'win32') {
+                return vendorPath.replace(/\\/g, '\\');
+            }
+
+            return vendorPath.replace(/\\/g, '/');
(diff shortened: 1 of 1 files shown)


matical:
Just an update, this also fails with a fresh L5.5 install (laravel-mix@1.4.5, npm@5.3.0, node@8.6.0).

None of the files throws 404, and seems to be returning the right ones?, but none of the extracted modules work. Tested with all the default included dependecies in the fresh install (Default example Vue component, jQuery etc don't work if extracted).

JeffreyWay:
I'd rather we normalize the URLs, rather than doing specific checks for windows like this PR.

MattyBalaam:
@JeffreyWay is there a link to where this might be fixed?

PR #1074, opened 2017-07-29
Add cross-env to support other environment

adityapurwa:
I added cross-env to support other environment - The command NODE_ENV=development (etc.) on the NPM scripts does not work on windows. Using cross-env helps resolve that issue.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -59,9 +59,9 @@ As a tip, consider adding the following NPM scripts to your `package.json` file,
 
 ```js
   "scripts": {
-    "dev": "NODE_ENV=development webpack --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js",
-    "watch": "NODE_ENV=development webpack --watch --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js",
-    "hot": "NODE_ENV=development webpack-dev-server --inline --hot --config=node_modules/laravel-mix/setup/webpack.config.js",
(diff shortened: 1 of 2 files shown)


ruchern:
This has been discussed before: https://github.com/JeffreyWay/laravel-mix/pull/927

adityapurwa:
@ruchern is #927 only for the documentation? My changes actually changed the `packages.json` to include `cross-env` as dev dependencies - I think laravel-mix should just work by default, without requiring the user to manually install `cross-env` (because that's also how Laravel work, just install and it should work without additional efforts).

ruchern:
@adityapurwa https://github.com/JeffreyWay/laravel-mix/commit/cbe6418bcc0fd6d1fd3a26530107fc134d04fc64#diff-b9cfc7f2cdf78a7f4b91a753d10865a2

@JeffreyWay has explained it very clearly in his commit message.

adityapurwa:
I just noticed that, thanks you.

ruchern:
@adityapurwa I am not trying to say that you are wrong. In fact, for Windows, `cross-env` is much needed. However, looking at the commit message, I think this is less likely to be merged.

adityapurwa:
@ruchern Yes, I agree with you, this request is likely not going to be merged after I noticed @JeffreyWay commit message, that's why I am closing this pull request - don't get me wrong, I really appreciate your effort on pointing @JeffreyWay message and I am thankful for that :)

Issue #1006, opened 2017-07-16
Allow modifying webpackConfig after smart merge


jbbr:
I experienced various problems when trying to modifying the webpack config with `mix.webpackConfig()`.

E.g adding options to loaders of predefined rules does not work because the predefined rule remains in the rules array:

```js
mix.webpackConfig({
 module: {
   rules: [{
    test: /\.s[ac]ss$/,
    loaders: [
      {loader: 'style-loader'},
      {loader: 'css-loader', options: {sourceMap: true}},
      {loader: 'sass-loader', options: {sourceMap: true}},
    ],
  }]
 }
})
```

When removing the `options` object from these loaders the predefined rule gets overwritten as expected.

This seems to be a bug in https://github.com/survivejs/webpack-merge

Nevertheless, it would be great to have the possibility of changing the real webpackConfig after the auto-build of laravel-mix.
An easy way to archive this would be to dispatch an event in the `build()` method of `WebpackConfig` and pass the reference to the webpackConfig object as data.

```js
build() {
        this.buildEntry()
            .buildOutput()
            .buildRules()
            .buildPlugins()
            .buildResolving()
            .mergeCustomConfig();
  
        Mix.dispatch('configReady', this.webpackConfig);
  
        return this.webpackConfig;
}
```

Modifying the config from the `Mix.listen` callback does work fine afterwards. Removing and replacing predefined rules is easy with `config.module.rules = config.module.rules.filter(rule => String(rule.test) !== String(/\.s[ac]ss$/))`

It would be great to have such an event or other way of modifying the whole object. Thanks!

PR #1077, opened 2017-07-30
Dispatch event after building webpack config

jbbr:
This implements #1006 

Allows the following syntax to freely modify the webpack config without the limitations of the `webpack-merge` smart merge.

Example:

```js
// webpack.mix.js

Mix.listen('configReady', (webpackConfig) => {

  // remove original rule
  webpackConfig.module.rules = webpackConfig.module.rules.filter(rule => String(rule.test) !== String(/\.s[ac]ss$/))

  // add modified rule
  webpackConfig.module.rules.push({
    test: /\.s[ac]ss$/,
    loaders: [
      {loader: 'style-loader'},
      {loader: 'css-loader', options: {sourceMap: true}},
      {loader: 'sass-loader', options: {sourceMap: true}},
    ],
  })

})
```

### DIFF
--- a/src/builder/WebpackConfig.js
+++ b/src/builder/WebpackConfig.js
@@ -27,6 +27,8 @@ class WebpackConfig {
             .buildResolving()
             .mergeCustomConfig();
 
+        Mix.dispatch('configReady', this.webpackConfig);
+
         return this.webpackConfig;
     }
 


jameshulse:
It could be handy to document this feature (I can't find it from googling) as I think it is a common use-case to want the final webpack config for posting bug reports.

PR #1097, opened 2017-08-04
Add browsersync support for dd function

### LINKED ISSUE #1093: BrowserSync does not refresh on dd()
- Laravel Mix Version: 1.4.2
- Node Version (`node -v`): v6.10.0 
- NPM Version (`npm -v`): 3.10.10
- OS:  Windows 8 64-bit

### Description:
The browserSync does not refresh if the page returns with dd() because the browserSync js was not inserted on page without ```<html><body></body></html>``` tags.
``` <script async="" src="/browser-sync/browser-sync-client.js?v=2.18.13"></script>```

### Steps To Reproduce:
```
// Controller
// No auto refresh when file change
public function index() 
{
  dd('test');
}
```

### PR DESCRIPTION
**Pull request by kennith**
browsersync support for dd function.

Close #1093 

Credit to @jnbn

### DIFF
--- a/src/builder/webpack-plugins.js
+++ b/src/builder/webpack-plugins.js
@@ -78,7 +78,15 @@ module.exports = function () {
                         'resources/views/**/*.php',
                         'public/js/**/*.js',
                         'public/css/**/*.css'
-                    ]
+                    ],
+                    snippetOptions: {
+                       rule: {
+                           match: /(<\/body>|<\/pre>)/i,
+                           fn: function (snippet, match) {
+                               return snippet + match;
+                           }
+                       }
+                    }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by jnbn** [comment]
Thanks @kennith

**Comment by mijgame** [comment]
Is there a way to disable this / make it configurable? Personally I don't always want dd() to auto reload on changes. I use it to inspect responses from an api for example while writing code to process that data, which auto-reload would make impossible.

**Comment by kennith** [comment]
@mijgame Add the rules in the snippetOptions in your browserSync config from webpack.mix.js to 
```
snippetOptions: {
  rule: {
    match: /<\/body>/i
  }
}
```

PR #1106, opened 2017-08-09
Update recommended npm-scripts to use local webpack version in docs

simonvomeyser:
Hey Jeffrey, had an issue with your recommended npm scripts in the docs: 

When not having webpack globally installed, there was an error because the global install was used.

Hope this was not of any purpose and thank you for your work :)

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -59,9 +59,9 @@ As a tip, consider adding the following NPM scripts to your `package.json` file,
 
 ```js
   "scripts": {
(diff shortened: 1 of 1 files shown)


ruchern:
You do not actually need webpack to be globally installed for the current `package.json` scripts to work.

DannyFeliz:
I think this should be merge

yansern:
@simonvomeyser npm look for bin files from your project folder first at `your_project_folder/node_modules/.bin/` before resorting to globals. You might need to check if the `.bin` folder exists in your setup.

PR #1154, opened 2017-09-01
Allow file-loader (image and font) directories to be customised - resolves #1090

### LINKED ISSUE #1090: Save fonts and images to specific folders
Hi, 

I am using the latest Webpack and Laravel-mix version. 

I am new to both :) 

I am able to save the js and sass files to the folders that I require but have not been able to set the specific folders that I need the images and fonts to save too. 

This is a non Laravel build and I am trying to use it on an existing project that does not use the file structure recommended. 

Could someone please point me in the right direction. 

Thank you

l1955454:
This change allows setting the output directories for assets processed by the the file-loader - at the moment only images and fonts:

```js
mix.options({
  fileLoaderDirs: {
    images: 'img',
    fonts: 'web-fonts'
  }
});
```

Previously you'd have to remove the existing rules and add a modified version by listening to `configReady` and modifying `webpackConfig` directly - too much boilerplate, and not update-proof.

I think these options are especially necessary because the `webpack.config.js` file isn't left in the project root.

Resolves #1090.

### DIFF
--- a/src/builder/webpack-rules.js
+++ b/src/builder/webpack-rules.js
@@ -72,10 +72,10 @@ module.exports = function () {
                 options: {
                     name: path => {
                         if (! /node_modules|bower_components/.test(path)) {
-                            return 'images/[name].[ext]?[hash]';
+                            return Config.fileLoaderDirs.images + '/[name].[ext]?[hash]';
                         }
 
-                        return 'images/vendor/' + path
+                        return Config.fileLoaderDirs.images + '/vendor/' + path
                             .replace(/\\/g, '/')
                             .replace(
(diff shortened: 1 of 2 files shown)


l1955454:
Would love for this to land before the next [tag](/JeffreyWay/laravel-mix/issues/1152).

l1955454:
This one is getting pretty popular @JeffreyWay, any indication if/when it'll be merged?

wouter2203:
It would be great if the `test` would be able to be customized as well, currently my `.svg` images are saved in the `/fonts/` directory.

I could of course just save them both to the `/images`  folder after this is merged, but it would be cleaner if I could just edit the `test` property.

Issue #1024, opened 2017-07-20
How to implicitly load global SCSS files in Vue Single File Components?


corysimmons:
- Laravel Mix Version: 1.2.2
- Node Version (`node -v`): v7.4.0
- NPM Version (`npm -v`): 5.2.0
- OS: macOS Sierra

### Description:

I'd like to be able to import a couple Sass files globally: https://vue-loader.vuejs.org/en/configurations/pre-processors.html#loading-a-global-settings-file

I can do this by explicitly defining the `@import` in each `.vue` file, but this is pretty brittle (if I move some `.scss` into a new directory my `@import`s will break).

Come to think of it, I probably won't be migrating these really generic `.scss` files much, but at this point I'm just curious. 

I tried:

```js
let mix = require('laravel-mix');
const path = require('path')

mix
  .js('resources/assets/js/app.js', 'public/js')
  .sass('resources/assets/sass/app.scss', 'public/css')
  .webpackConfig({
    module: {
      rules: [
        {
          test: /\.vue$/,
          exclude: /(node_modules|bower_components)/,
          loader: 'sass-resources-loader',
          options: {
            resources: path.resolve(__dirname, 'resources/assets/sass/_variables.scss')
          }
        }
      ]
    }
  })
```

...per:

- https://laravel.com/docs/5.4/mix#custom-webpack-configuration
- https://vue-loader.vuejs.org/en/configurations/pre-processors.html#loading-a-global-settings-file

...but kept running into:

```fish
 error  in ./resources/assets/js/components/Foo.vue

Module build failed:
undefined
             ^
      Undefined variable: "$fav-color".
      in /Users/yo/Desktop/Blah/resources/assets/js/components/Foo.vue (line 89, column 15)
```

(`$fav-color` exists in `_variables.scss`)

I suspect the problem lies somewhere in your [webpack-rules.js](https://github.com/JeffreyWay/laravel-mix/blob/29edf9ccbb73bf2048364a5890700365236395c9/src/builder/webpack-rules.js), but haven't really tinkered with it.

Feel free not to worry about this, but it might be nice to add [vue-loader's official solution](https://vue-loader.vuejs.org/en/configurations/pre-processors.html#loading-a-global-settings-file) to Mix.

Thanks for all your hard work. Grats on the kiddo.

JeffreyWay:
Closing this in favor of the PR. https://github.com/JeffreyWay/laravel-mix/pull/1163

PR #1163, opened 2017-09-05
#1024 Add globalVueStyles configuration option

NoelDeMartin:
Proposal to solve issue #1024

Having a global sass file with variables, mixins and functions is a common scenario, and right now the only option to use such file in .vue component files is importing it manually every time. From the vue-loader documentation we can see how they recommend solving this: [Loading a global settings file](https://vue-loader.vuejs.org/en/configurations/pre-processors.html#loading-a-global-settings-file)

Having all this into account, I think it'd be sensible to have such an option in the mix manifest.

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -15,6 +15,7 @@ mix.options({
 A handful of Mix options and overrides are available, should you require them. Please take note of the options above, as well as their default values. Here's a quick overview:
 
 - **extractVueStyles:** Extract `.vue` component styling (CSS within `<style>` tags) to a dedicated file, rather than inlining it into the HTML.
+- **globalVueStyles:** Indicate a file to include in every component styles. This file should only include variables, functions or mixins in order to prevent duplicated css in your final, compiled files. This option only works when `extractVueStyles` is enabled and requires [sass-resources-loader](https://github.com/shakacode/sass-resources-loader) package to be installed.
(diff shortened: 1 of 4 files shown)


AustinPaquette:
Is there a way I can use this right now? It's an absolute gem of an addition.

NoelDeMartin:
@AustinPaquette Until @JeffreyWay accepts or rejects the pull request, you could either use a fork of laravel-mix or implement this on the webpack config of your app.

Even thou I made this, I am still importing a sass file in each component file. It's a drag having to repeat that every time but it's maintainable at the moment. Depending on jeffrey's opinion on this I may start using a fork myself.

AustinPaquette:
Yeah, I'll wait this out for a few more days but if it's closed or becomes stagnant I'll do that. Super useful feature for sure!

JeffreyWay:
Thanks! Sorry for taking so long to getting around to this PR.

AustinPaquette:
> Thanks! Sorry for taking so long to getting around to this PR.

AustinPaquette:
Already got it working on my side project just so I could test it out.

@NoelDeMartin you're my hero of the week. Happy Monday to you good sir.

NoelDeMartin:
Glad to help :D Thanks Jeffrey!

Issue #1194, opened 2017-09-16
Bump vue-loader version to 13.x


ankurk91:
- Laravel Mix Version: 1.4.3 (`npm list --depth=0`)
- Node Version (`node -v`): 6.11
- NPM Version (`npm -v`): 4.6
- OS: Ubuntu 16

### Description:
Vue-loader 13 was released on june 29
https://github.com/vuejs/vue-loader/releases/tag/v13.0.0
but laravel-mix still using v12.x

PR #1206, opened 2017-09-19
Update vue-loader to 13.0.5.

yansern:
`js-beautify` broke and `vue-loader` switched to prettier. #1194 

See:
https://github.com/beautify-web/js-beautify/issues/1247
https://twitter.com/vuejs/status/909785891941617664

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "style-loader": "^0.18.1",
     "uglify-js": "^2.8.28",
     "uglifyjs-webpack-plugin": "^0.4.6",
-    "vue-loader": "^12.1.1",
+    "vue-loader": "^13.0.5",
     "vue-template-compiler": "^2.4.2",
     "webpack": "^3.5.0",
     "webpack-chunk-hash": "^0.4.0",


ruchern:
Did you try checking if this will break anything else? Going from v12 to v13 might break something thus I did not really want to bump the version earlier.

yansern:
I have that concern myself too. But I've tested building my vue files on laravel with v13 and it went well. Replace laravel-mix in your package.json to `"laravel-mix": "yansern/laravel-mix",` and see if yours compile fine?

ThomHurks:
After updating to Laravel-Mix 1.4.4 I suddenly get this Vue error: ```vue.common.js:478 [Vue warn]: Failed to mount component: template or render function not defined.``` and none of my single-file components work anymore. I would assume it's caused by the vue-loader change. Reverting to 1.4.3 fixes my issue and Vue works fine. Any idea what's wrong?

yansern:
@ThomHurks If you use `require('foo.vue')` you'll need to change it to `require('foo.vue').default`. See: https://github.com/vuejs/vue-loader/releases/tag/v13.0.0

I just realized I use `import` instead of `require` when loading vue files that's why I wasn't getting any errors.

@JeffreyWay I think I need to submit another pull request, adding `esModule: false` on the vue-loader options on the the `webpack-rules.js` file for backwards compatibility. See: https://github.com/vuejs/vue-loader/blob/master/docs/en/options.md#esmodule

ThomHurks:
Ah, I'm using ```import``` for Vue and Vue plugins, but I'm using ```require``` for my own components. I'll try switching to using ```import``` or ```require().default``` for those. Probably a good idea to put this in the release notes :)

JeffreyWay:
Ah yeah this is a breaking change. Need to fix that. @yansern can you submit that second PR?

yansern:
@JeffreyWay Done. Hope this works!

PR #1209, opened 2017-09-19
Add esModule: false to vue-loader options for compatibility with vue files loaded using require().

yansern:
Add esModule: false to vue-loader options for compatibility with vue files loaded using require(). Related: #1206


yansern:
@ThomHurks Good idea! Just created a pull request for that. esModule option can be enabled using:
```
mix.options({
  vue: {
    esModule: true
  }
});
```

PR #1211, opened 2017-09-20
Make esModule option configurable.

yansern:
For devs who want the esModule option, do this in your `webpack.mix.js` file.
```
mix.options({
  vue: {
    esModule: true
  }
});
```
Related: #1209


molerat619:
What does esModule do?

PR #1214, opened 2017-09-20
Use smartStrategy() in .webpackConfig method

dragermrb:
This allows to use strategies with mergeStrategy, for example if you need to prepends some plugin:

```
mix.webpackConfig({
    plugins:[ new ngAnnotatePlugin({add: true}) ] // My custom config
}, {
    plugins: 'prepend' // Plugins strategy
});

```
This way **ngAnnotatePlugin** will run before **UglifyJsPlugin**, which is recommended in AngularJS applications.


JeffreyWay:
Because you can call `mix.webpackConfig()` multiple times, I'd rather not lock the strategy type for every use. 

My instinct is to close this for now, since I can't think of anyone else who has requested this particular functionality.

PR #1225, opened 2017-09-29
Update installation.md

kevdotbadger:
Update instructions to create the src path and the app.js/app.scss files.


JeffreyWay:
I think it's okay as is. But thanks! :)

Issue #1235, opened 2017-10-02
mix.version() raise error


EmranAhmed:
- Laravel Mix Version: #.#.# (`npm list --depth=0`)

```
├── cross-env@5.0.5
└── laravel-mix@1.4.5
```

- Node Version (`node -v`):  `v8.6.0`

- NPM Version (`npm -v`): `5.3.0`
- OS: OSX El Capitan ( 10.11.6 )

### Description:

My `webpack.mix.js` file contents:

```js
const mix      = require('laravel-mix');
const min      = Mix.inProduction() ? '.min' : '';

mix.version();
// mix.sourceMaps();

 mix.js(`src/index.js`, `dist/index.js`);

```
When I add `mix.version();` it raise error, but if i remove `mix.version();` its works fine :)

Error when `mix.version();` used:

```js
fs.js:653
  return binding.open(pathModule._makeLong(path), stringToFlags(flags), mode);
                 ^

Error: ENOENT: no such file or directory, open '/dist/index.js'
    at Object.fs.openSync (fs.js:653:18)
    at Object.fs.readFileSync (fs.js:554:33)
    at File.read (/Users/xxx/node_modules/laravel-mix/src/File.js:180:19)
    at File.version (/Users/xxx/node_modules/laravel-mix/src/File.js:190:25)
    at Manifest.hash (/Users/xxx/node_modules/laravel-mix/src/Manifest.js:55:65)
    at manifest.forEach.file (/Users/xxx/node_modules/laravel-mix/src/plugins/CustomTasksPlugin.js:79:47)
    at Array.forEach (<anonymous>)
    at CustomTasksPlugin.applyVersioning (/Users/xxx/node_modules/laravel-mix/src/plugins/CustomTasksPlugin.js:79:18)
    at Compiler.compiler.plugin.stats (/Users/xxx/node_modules/laravel-mix/src/plugins/CustomTasksPlugin.js:12:22)
    at Compiler.applyPlugins (/Users/xxx/node_modules/tapable/lib/Tapable.js:61:14)
    at Watching._done (/Users/xxx/node_modules/webpack/lib/Compiler.js:104:17)
    at compiler.emitRecords.err (/Users/xxx/node_modules/webpack/lib/Compiler.js:78:19)
    at Compiler.emitRecords (/Users/xxx/node_modules/webpack/lib/Compiler.js:375:38)
[...]

PR #1236, opened 2017-10-03
Fix minor versioning issue

talvbansal:
When versioning files sometimes the files path would appear to be a relative path prefixed with / causing an error in locating the file to be versioned.

EG : #1235 

This small fix checks to see if a file's path starts with mix's root path and if it doesn't it prefixes it so that an absolute path can be created and the real file can be found by the fs module.

Also added vue to the package.json file as without it the test suite was failing.


JeffreyWay:
I believe this was fixed in a separate PR that we've already merge, so I'll give this a close.

PR #1243, opened 2017-10-04
sourceMaps type style as param

dsolanor:
new param added to sourceMap function to configure source mapping style

### DIFF
--- a/src/Api.js
+++ b/src/Api.js
@@ -357,10 +357,9 @@ class Api {
      * Enable sourcemap support.
      *
      * @param {Boolean} productionToo
+     * @param {string} type
      */
-    sourceMaps(productionToo = true) {
-        let type = 'inline-source-map';
-
+    sourceMaps(productionToo = true, type = 'inline-source-map') {
         if (Mix.inProduction()) {
             type = productionToo ? 'cheap-source-map' : false;
         }


joinsito:
Very useful change to be able to specify the sourcemap style

JeffreyWay:
Thanks!

lukepolo:
@JeffreyWay Could you tag a new version with this included?

ellisio:
@JeffreyWay I know this is merged, but this logic is confusing me in terms of allowing the developer flexibility. Say I want to use another source map type in production, just because I want to, this method is not flexible enough to allow me to do that. Instead I have to go through `webpackConfig` and set the source map there.

Is there a specific reason why you have a `Mix.inProduction()` check within this method? Maybe it would be better to remove that check, default to `cheap-source-maps`. Maybe like this:

```
    sourceMaps(productionToo = true, type = 'cheap-source-map') {
        if (! productionToo && Mix.inProduction()) {
            type = false;
        }

        Config.sourcemaps = type;

        return this;
    };
```

This, to me, reads better because:

```
Mix.sourceMaps(); // Uses "cheap-source-map" in dev and production.
Mix.sourceMaps(false); // Uses "cheap-source-map" in dev only.
Mix.sourceMaps(true, 'inline-source-map'); // Uses "inline-source-map" in dev and production.
Mix.sourceMaps(false, 'inline-source-map'); // Uses "inline-source-map" in dev only.
```

Am I missing something further into the core for why you have it the way it is?

ellisio:
@JeffreyWay Or maybe better yet, like this?

```
    sourceMaps(type = 'cheap-source-map', productionToo = true) {
        if (! productionToo && Mix.inProduction()) {
            type = false;
        }

        Config.sourcemaps = type;

        return this;
    };
```

Because then it would be configurable like this:

```
Mix.sourceMaps(false); // Disabled in dev and prod.
Mix.sourceMaps(); // Uses "cheap-source-map" in dev and prod.
Mix.sourceMaps('cheap-source-map', false); // Uses "cheap-source-map" in dev only.
Mix.sourceMaps('inline-source-map'); // Uses "inline-source-map" in dev and prod.
Mix.sourceMaps('inline-source-map', false); // Uses "inline-source-map" in dev only.
```

Then if someone wants to use a different source map in dev, than in production, have them to that in the `webpack.mix.js` like so:

```
if (Mix.inProduction()) {
    Mix.sourceMaps(); // Use "cheap-source-map" in production.
} else {
    Mix.sourceMaps('inline-source-map'); // Use "inline-source-map" everywhere else.
}
```

Just some thoughts.

stephan-v:
I agree with @ellisio I am working on setting up Sentry right now for error reporting and this bit of code made me chuckle because it seems little thought went into this.

You need to be able to specify different sourcemaps depending on your environment.

Issue #1000, opened 2017-07-14
Feature Request: Config var for dev server host / port


eheaton:
- Laravel Mix Version: 1.2.1
- Node Version: 6.10.3
- NPM Version: 5.2 
- OS: Mac OS Sierra

### Description:
 It'd be great to be able to set the host / port on the WebPack dev server, via. a config option. You can partially do it with the `devServer` WebPack config options, but the inline request string is hard coded in the WebPackConfig

WebPackConfig.js:68
```
        this.webpackConfig.output = {
            ...
            publicPath: Mix.isUsing('hmr') ? (http + '://localhost:8080/') : ''
            ...
        };
```

ankurk91:
@eheaton
Then laravel blade helper **mix()** also needs to be configurable.

JeffreyWay:
If this is needed, can someone submit a PR to make the localhost:8080 bit a config option in src/Config.js?

sifex:
Added PR for this, 
https://github.com/JeffreyWay/laravel-mix/pull/1246

kylekatarnls:
@JeffreyWay Yes this is usefull for running multiple Laravel instance at the same time.

PR #1246, opened 2017-10-05
Support for HMR on different hosts and ports

sifex:
Fixes https://github.com/JeffreyWay/laravel-mix/issues/1000

Addition of host and port options in Laravel Mix and passing them into Webpack Dev Server. This allows both external viewing of hot reloading (which works amazing) and also allows the changing of port in case 8080 is taken on the host machine (which it was for me)

One thing to note is the hijacking of the `public/hot` file to be able to communicate with the Laravel mix helper. Instead of writing "hot reloading", I've changed it to be simply "http/s" plus the new host and port.

Tested without port and host variables, and as expected they default back to localhost:8000.

---

Important to note, the Laravel Mix Helper in the Laravel Framework needs to change, should be adding a related PR fairly shortly.

Edit: https://github.com/laravel/framework/pull/21545 PR Added


kylekatarnls:
Hi, there is 1 space in excess (`let http...` is currently shifted) if you want to fix it in the same time with this PR that I would really enjoy to see merged. :)

sifex:
Good catch @kylekatarnls

docs/hot-module-replacement.md
```diff
@@ -42,6 +42,20 @@ However, it can be a burden to manually change this URL for production deploys.
...
+```js
+mix.options({
+    hmrOptions: {
+        host: '10.1.1.2',
+        port: '3030'
+    }
+});
+```
+
+Important: Currently, support for the `{{ mix('js/bundle.js') }}` blade helper is pending PR.
```
nmfzone: I think this is not relevant right now.

sifex: The PR request?

nmfzone: Yes, `mix()` already get merged, right?

nmfzone: Ah sorry, I don't get it. I just realize it's your own PR.

sifex:
Thanks @JeffreyWay  :)

dmason30:
@sifex It appears the docs you added were removed, I think because laravel hadn't released the `mix` helper change yet, but they released it in 5.5 so the docs should be re-added?

Issue #1253, opened 2017-10-11
Update ISSUE_TEMPLATE.md . npm cache clear throws error


acacha:
- Laravel Mix Version: 1.5.0
- Node Version (`node -v`): v6.10.2
- NPM Version (`npm -v`): 5.4.2
- OS: Ubuntu

### Description:

Steps provided by new issue template throws an error. Change npm cache clear with npm cache verify

### Steps To Reproduce:

Execute:
```
rm -rf node_modules && npm cache clear && npm install
npm ERR! As of npm@5, the npm cache self-heals from corruption issues and data extracted from the cache is guaranteed to be valid. If you want to make sure everything is consistent, use 'npm cache verify' instead.
npm ERR! 
npm ERR! If you're sure you want to delete the entire cache, rerun this command with --force.

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/sergi/.npm/_logs/2017-10-11T15_41_26_165Z-debug.log
```

ankurk91:
```
npm cache clear -f
```
Above command should work on npm 5

acacha:
Yes but then there is a warning about not necessary to clear cache. Maybe verify is better?

ankurk91:
From the npm [docs](https://docs.npmjs.com/cli/cache)

>verify: Verify the contents of the cache folder, garbage collecting any unneeded data, and verifying the integrity of the cache index and all cached data. 

But here we want user to install latest versions of all dependencies before reporting any issue.

Verify option does not fit here.

PR #1254, opened 2017-10-11
Update ISSUE_TEMPLATE.md

acacha:
See https://github.com/JeffreyWay/laravel-mix/issues/1253


acacha:
Or maybe use -force option?

JeffreyWay:
I'm not sure we need this.

PR #1255, opened 2017-10-11
Update to avoid errors executing npm cache clear

acacha:
New versions of node throws an error with npm cache clear. Use verify instead


JeffreyWay:
I don't think we need verify here? I'll check to see why node is throwing an error.

PR #1259, opened 2017-10-12
Fix failing test by adapting it to new conditions

tpraxl:
The test failed, because there were two entries
'transform-object-rest-spread' in the options.plugins array.

This has most likely to do with src/config.js which declares
'transform-object-rest-spread' as a default plugin.


JeffreyWay:
Thanks, @tpraxl!

PR #1261, opened 2017-10-13
Make project work out of the box without laravel

dpollard:
Before this change, on windows machines, the code would throw an error out of the box. With this change, everything works smoothly as it should, and can be used on all platforms without needing to search for the fix.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -109,7 +109,7 @@ module.exports = function () {
          *
          * @type {String}
          */
-        publicPath: '',
+        publicPath: 'public',
 
 
         /**


ruchern:
However, this package is primarily used for Laravel. Users should be given the choice to edit the path which they already do.

dpollard:
But this changes nothing for laravel users anyway (especially if they already change it) and it ensures another use case out of the box.

ruchern:
Okay, I have just tested this change and it worked with my out of the box `Laravel`.

dpollard:
Yep, and this change would make the project work out of the box without laravel on windows machines.

JeffreyWay:
We can't assume that public directory for standalone projects, when we have no idea what your directory structure looks like.

What error do you get? You need to use mix.setPublicPath() to specify this.

artemsky:
Just use `mix.setPublicPath()` if you have custom project.
This one is created for laravel apps.
This change will break 90% of projects

PR #1267, opened 2017-10-13
Fix tests for win32

artemsky:
There are several problems with laravel-mix and it's testing on Windows related to backslashes in paths.
**This should fix 99.99% of issues.**

#### Just one left.

Its related to `manifest.js` generation where we have following situation...

Imgur
And it's not a surprise that md5 hashes are not identical.

@JeffreyWay How do you think, maybe we should inject [slash](https://www.npmjs.com/package/slash)  into laravel-mix and normalize all paths according to *nix systems format or to do it in some other efficient way?


JeffreyWay:
I'll try to go through this soon. Sorry for taking so long. :)

ankurk91:
@artemsky 
Can you rebase with upstream?

MohannadNaj:
Great @artemsky !

running win7-64 , I was getting countless failures, mostly -as you mentioned- due to the backslash issue.

this PR will make all the tests passing **EXCEPT this**:

```
1 test failed [02:29:52]
  file » it can minify JS files.
  C:\nodes\laravel-mix\node_modules\mock-fs\lib\binding.js:415

  Error thrown in test

  Error:

    [Error: ENOENT, no such file or directory '\\?\C:\nodes\laravel-mix\
                var one = 'one';
                var two = 'two';
            ']
```

I think the blame is on src\File.js `minify()` method, used to call `uglify.minify` with the argument: `file.path()`, but then replaced to pass the content instead: `file.read()`. diff here:

https://github.com/JeffreyWay/laravel-mix/pull/1267/files#diff-182a1c65ef29ba6e05eee122c71511e6L221

while our `uglifyCss.minify` accepts the content as the first argument if we are in the css context, `UglifyJS.minify` accepts paths only.

artemsky:
@ankurk91 done

kohenkatz:
This should probably be closed in favor of #1425, which is more up to date and doesn't require adding another library.

JeffreyWay:
We went with a slightly different PR/approach. I think this is fully fixed now. Will tag this week.

Thanks! And sorry for delay.

PR #1273, opened 2017-10-16
Upgrade vue-template-compiler version

ahmed-aliraqi:


ruchern:
I have explained why the version is set as `^2.0.0` in #1264.

medz:
@ruchern I think this PR is excellent, or dependent version can be modified to `^ 2. *`. This is compatible with the 2.5.x version of Vue.js.

ruchern:
@medz https://docs.npmjs.com/files/package.json#dependencies

Look at this documentation.

Quoting from this PR, each time `vue` is updated, the `vue-template-compiler` will only be pulling in `2.5.2` as the highest version.

The value that I have set is `^2.0.0`, so whenever `vue` gets an update `vue-template-compiler` will also get updated to match the version on `vue`. This should allow `vue-template-compiler` to pull in the latest package based off `2.0`

e.g. Vue got updated to `2.6.0`

With my previous PR, `vue-template-compiler` will automatically pull in `2.6.0` which is by right the latest version from the vue organisation side.

With this PR, `vue-template-compiler` will only pull in `2.5.2` instead of `2.6.0`, which has no change whenever `vue-template-compiler` is being updated.

PR #1274, opened 2017-10-16
Update package-lock.json

ahmed-aliraqi:


JeffreyWay:
I think the dependency chain has probably changed since submitting this, so I'll go ahead and close and push an updated package-lock. Thanks!

Issue #1266, opened 2017-10-13
Laravel Mix empties file while “watch”-ing


OurBG:
- Laravel Mix Version: 0.12.1
- Node Version 8.1.2
- NPM Version (`npm -v`): 5.4.2
- OS: CentOS

### Description:

I am running npm run watch and it works perfectly with all my SASS or CSS files. I also have this line of code there:

`.scripts(['resources/assets/js/specific/dashboard.js'], 'public/js/specific/all.js')`

Every time I run npm run watch dashboard.js is succesfully compiled into all.js.

However every time I make a change in dashboard.js laravel mix instantly empties all.js never to fill it up again. If I Ctrl+c the process and run npm run watch again it compiles it successfully again.

This is bad because every time I make a change in the js (that I'm currently developing) I need to go back and restart the process and wait for it.

JeffreyWay:
Please update your install of Laravel Mix to a current version. You have a very outdated version.

OurBG:
Your readme File:

> Laravel Project
> 
> Laravel 5.4 ships with everything you need to get started. Simply:
> 
> Install Laravel
> Run npm install
> Visit your webpack.mix.js file, and get started!

I have installed Laravel 5.4 and somehow I am running a **very** outdated version...

ruchern:
Laravel Mix. It’s now 1.5.1.

OurBG:
I know ... My point was that Laravel-Mix readme says "Laravel 5.4 ships with everything you need" which apparently is not true since it comes with *very* outdated version.

Should npm update laravel-mix update it?

devcircus:
Just sent a PR to the docs to fix the reference to Laravel 5.4. #1285

PR #1285, opened 2017-10-20
remove reference to Laravel 5.4

devcircus:
## Before
The installation docs refer to Laravel 5.4:
> Laravel 5.4 ships with everything you need to get started. Simply:

## After
Instead of having to keep this updated, I just dropped the version number.
> Laravel ships with everything you need to get started. Simply:


devcircus:
In #1266, a user had installed 5.4 based on this, which of course led to having an outdated version of mix.

Issue #1231, opened 2017-10-01
Why do SVG images compile into the fonts folder?


Spazer9:
I'm wondering why this behavior happens and how to override it?

Regular images (png, jpg) go into the correct `images` folder. but SVG's all go to the `fonts` folder.

My css:
```
.foo{ background: url(../images/normal_icon.png) }
.bar{ background: url(../images/icon.svg) }
```
Converts to:
```
.foo { background: url(/image/normal_icon.png?id=a72845c0a452ccc7003a) }
.bar { background: url(/fonts/icon.svg?id=085742210e6d227f68b0) }`
````

My mix file is:

```
const { mix } = require('laravel-mix')

mix.setPublicPath('public')
   .js('app/Assets/js/app.js', 'js')
   .sass('app/Assets/css/test.scss', 'css')
   .copy('app/assets/images', 'public/images', false)
   .version()
   .version('public/images/**/*')
   .disableNotifications();
```

ruchern:
Related issue #214

stephan-v:
Time to grab a bucket of popcorn until this finally gets resolved.

ruchern:
@stephan-v Haha, yeah. I guess this is becoming a thing and your PR should be a good fix to this.

Mdejonge:
I had the same problem, but finally fixed it. You are using `../images/normal_icon.png` as path. Try removing the dots at the beginning of the path, so it looks like this: `/images/normal_icon.png`. 
For some reason it will compile all the paths with `../` to the fonts folder.

giovannipds:
@Mdejonge I wouldn't say that's a "solution", it affects how we read and write url paths. BUT, it was important that you've mentioned.

A solution would be a PR changing the webpack config. =)

heihachi88:
Any news on this?

stephan-v:
@heihachi88 

I had a simple fix for this but it go closed.

https://github.com/JeffreyWay/laravel-mix/pull/672

I simply can not see why this would not go through now, since there is an overwhelming response to drop SVG's from being seen as fonts.

adriandmitroca:
FYI: SVG fonts are deprecated and useless anyway, really don't see any point of having this -
 https://caniuse.com/#feat=svg-fonts

@JeffreyWay

stephan-v:
@adriandmitroca 

I would either like to see it removed or receive a better explanation from @JeffreyWay than:

> "This would break everything though."

[...]

Issue #214, opened 2017-01-28
copied svg asset broken


OussamaElgoumri:
i'am using flag-icon-css, and loaded by mix, some of the svg assets copied by mix to public/fonts are broken, but when copied manuelly work.

This is my mix config:

``` javascript
const { mix } = require('laravel-mix');

mix.disableNotifications();

mix
    .js('resources/assets/js/welcome/welcome-vendor.js', 'public/js')
    .js('resources/assets/js/welcome/welcome.js', 'public/js')
    .sass('resources/assets/sass/welcome/welcome-vendor.scss', 'public/css')
    .sass('resources/assets/sass/welcome/welcome.scss', 'public/css')
    .sourceMaps()
    .version();
```

is there a way to tel me mix to not copy these dependencies, or to use a custom shell script to copy dependencies?

Thank you for `mix.disableNotification()` :)

JeffreyWay:
I'd need to see the related CSS, but any url that is absolute (/svg/path.svg) won't be modified by Webpack.

OussamaElgoumri:
This is the current mix configuration i'am using

```

mix.disableNotifications();

mix

    .js('resources/assets/js/channel-show/app.js',           'public/js/channel-show.js')
    .js('resources/assets/js/channel-show/vendor.js',        'public/js/channel-show-vendor.js')
    .js('resources/assets/js/channels-by-country/app.js',    'public/js/channels-by-country.js')
    .js('resources/assets/js/channels-by-country/vendor.js', 'public/js/channels-by-country-vendor.js')
    .js('resources/assets/js/embed/app.js',                  'public/js/embed.js')
    .js('resources/assets/js/embed/vendor.js',               'public/js/embed-vendor.js')
    .js('resources/assets/js/welcome/app.js',                'public/js/welcome.js')
    .js('resources/assets/js/welcome/vendor.js',             'public/js/welcome-vendor.js')

    .sass('resources/assets/sass/channel-show/app.scss',           'public/css/channel-show.css')
    .sass('resources/assets/sass/channel-show/vendor.scss',        'public/css/channel-show-vendor.css')
    .sass('resources/assets/sass/channels-by-country/app.scss',    'public/css/channels-by-country.css')
    .sass('resources/assets/sass/channels-by-country/vendor.scss', 'public/css/channels-by-country-vendor.css')
    .sass('resources/assets/sass/embed/app.scss',                  'public/css/embed.css')
[...]

PR #1337, opened 2017-11-23
The people have spoken - SVG needs to be removed

stephan-v:
https://github.com/JeffreyWay/laravel-mix/issues/1231
https://github.com/JeffreyWay/laravel-mix/pull/672
https://github.com/JeffreyWay/laravel-mix/issues/350
https://github.com/JeffreyWay/laravel-mix/issues/550
https://github.com/JeffreyWay/laravel-mix/issues/214

I can go on and find more but I doubt that is necessary. 

Nothing will break due to this change and I do not see how at this point with more and more complaints coming in that this can be ignored any longer.

### DIFF
--- a/src/builder/webpack-rules.js
+++ b/src/builder/webpack-rules.js
@@ -95,7 +95,7 @@ module.exports = function () {
 
     // Add support for loading fonts.
     rules.push({
-        test: /\.(woff2?|ttf|eot|svg|otf)$/,
+        test: /\.(woff2?|ttf|eot|otf)$/,
         loader: 'file-loader',
         options: {
             name: path => {


ankurk91:
Will wepack still pick SVG files referenced in source code?
Some people might be using SVG as images not only as fonts.

stephan-v:
Update your mix.webpackConfig. If you want case-specific packages to work and have their .svg fonts resolved you should create a rule for it. Not the other way around.

Bootstrap 4 dropped the SVG fonts for a good reason and there are plenty of ways to work around font-awesome.

The default should be not to treat .svg files as fonts. I should not have to create counter-rules simply because some packages(who are all moving away from svg fonts mind you) still use them.

I'm happy to hear about legitimate concerns but so far I see none.

ruchern:
@stephan-v There will be potential problems with Font Awesome. Looking at the v5 milestone, it does not seem like they are going to drop the SVG fonts for some reason.

stephan-v:
There is no problem if you extend your webpack config like I said before, if you want to deal with it expand upon the existing config.

If you really do not want to tackle this issue then atleast allow for the API to opt out of svg fonts.

ruchern:
@stephan-v I am not saying that we should not go ahead with this but since the depreciation of SVG fonts, Font Awesome should follow.

However, in order for Jeffrey Way to start implementing this into the package, these packages like Font Awesome have to comply first otherwise there would be too much work to do the checks in the code itself.

ankurk91:
I also wants to get rid of `svg` fonts but what if i am calling svg image file in my code somewhere.
```js
let logo = require('./img/company-logo.svg');
```
We should have a loader to handle these files.
Consider moving `svg` to img loader.

stephan-v:
Ok I hear you if that is the case. 

I will see if I have time to work on a solution which should make it easier to opt out of the SVG rule.

stephan-v:
> However, in order for @JeffreyWay to start implementing this into the package, these packages like Font Awesome have to comply first otherwise there would be too much work to do the checks in the code itself and to deconflict for the users is another story.

Again like I said it will be **your job** to override your webpack config and opt in to use SVG's as fonts. This should not be the default case. This would require no code re-work here, just in your local project. As it should be, since you want to do something that falls out of the default norm.

SVG fonts are being deprecated and I can supply the dozens of comments again who are in favor of dropping this but I would rather not.

Moving `svg` to the img loader is fine by me.

jm-mwi:
Where is the info that SVG fonts are being removed? The only references I found (with an admittedly short search) were from 10+ years ago @ Mozilla.

I'm not taking a stance, just looking for the references to the fact that SVG fonts are gone.

ankurk91:
@jm-mwi 
https://caniuse.com/#search=svg%20fonts

tpraxl:
+1 for removing svg fonts or at least do something that matches most use cases:

```js
// images
test: /(\.(png|jpe?g|gif)$|im[a]?g.*\.svg$)/,
// fonts
test: /(\.(woff2?|ttf|eot|otf)$|font.*\.svg$)/,
```
I was very surprised that this behavior is still the default in mix. :/

ruchern:
@tpraxl a PR will be good for this.

tpraxl:
@ruchern I would love to make a PR, but I can't manage to let the tests pick up on URLs in my scss fixture. So I'm currently unable to write a test for this, but I'll try to figure it out.

tpraxl:
https://github.com/JeffreyWay/laravel-mix/pull/1367

The mix manifest didn't reflect the resolved image / fonts, so I struggled for some time to write the test. 
Now, I really hope that this reflects most of the use cases and will make it into laravel-mix.

JeffreyWay:
One of these two PRs will be merged for the next major release. Will try to have that released before the end of the year.

domstubbs:
It would be so helpful to be able to provide an array of overrides for each of the default rules’ tests and excludes, at least then there would be a simple solution where people have differing use cases. I appreciate that it is supposed to be possible to do that kind of thing via webpackConfig() but I’ve spent ages trying to just move SVGs from the font loader to the image loader and it doesn’t seem to be possible without modifying webpack.config.js as well. Even then, if you want to largely retain the default config it’s pretty awkward.

FWIW, I was able to override one of the simpler default rules using webpackConfig() in the past and from looking around I think this might be a [limitation of webpack-merge](https://github.com/survivejs/webpack-merge/issues/89) with function type configs. I think that would explain why I could override the JS rule but not the font one.

JeffreyWay:
Fixed in #1367.

Issue #1338, opened 2017-11-23
Node-sass package is out of date


hoquangthaiholy:
Recently, when I try to install laravel-mix. It raises an error because of version 4.5.3 is out of date. Maybe, we need to change dependence on node-sass in package.json:

from
`"node-sass": "4.5.3"`
to
`"node-sass": "^4.5.3",`

Is it right?

georgewritescode:
Happened to me too, gives a 404 error when trying to download the binary for

https://github.com/sass/node-sass/releases/download/v4.5.3/win32-x64-59_binding.node

However, if we change it to the latest binary it works fine:
https://github.com/sass/node-sass/releases/download/v4.7.2/win32-x64-59_binding.node

Problem with the hard coding it seems

ruchern:
@georgewritescode Just manually do `npm install -D node-sass@latest` if you have to.

PR #1334, opened 2017-11-21
Update node-sass version in package.json

axtg:
Old version listed of node-sass (4.5.3) is no longer available at its known remote address, causing a clean npm install to break as laravel-mix still refers to it. Updating to 4.7.2 (as done for Bootstrap 4 beta 2), should solve this.

PR #1315, opened 2017-11-10
Update package.json

lordroseman:
After npm install on laravel55 project I got this error.. and I can't install laravel-mix even I download the node-sass seperately..,

````
Downloading binary from https://github.com/sass/node-sass/releases/download/v4.5.3/win32-x64-59_binding.node
Cannot download "https://github.com/sass/node-sass/releases/download/v4.5.3/win32-x64-59_binding.node":

HTTP error 404 Not Found
````

Upon checking there is no /win32-x64-59_binding.node in v4.5.3, so I'm proposing to update the node-sass to latest release of v4.6.0

Issue #1306, opened 2017-11-07
Allow node-sass 4.6.0?


roddypratt:
Any chance that the node-sass dependency could be updated to allow 4.6.0?

I seem to have shot myself in the foot with by updating to node.js 9.1 -  which 4.5.3 doesn't have a prebuilt binary for.  So (being on windows) I'm into Windows node-gyp build hell...

ruchern:
Have you tried manually install `node-sass` into your `devDependencies`?

roddypratt:
Hmm. I did `yarn add node-sass --dev` which runs fine and adds `"node-sass": "^4.6.0"` to package.json. but when I then add laravel-mix, it still fails downloading `https://github.com/sass/node-sass/releases/download/v4.5.3/win32-x64-59_binding.node` (which doesn't exist).

ruchern:
What about adding `Laravel Mix` then `node-sass`? Does that not work?

roddypratt:
Laravel Mix explicitly requires node-sass 4.5.3 according to it's package.json, so when you do `yarn add laravel-mix` yarn tries to download the correct pre-built binary for node-sass, according to your node.js version. I have node.js 9 installed, and there's no node-sass 4.5.3 binary for that version.

roddypratt:
just seen this commit message: https://github.com/JeffreyWay/laravel-mix/commit/03ae69849ac6c7454e299f964a9f977eab7e36c8

So looks like the 4.5.3 dependency is intentional right now, and Jeffrey is aware :)

roddypratt:
@ruchern @JeffreyWay Looks like node-sass 4.6.1 has fixed the watcher bug.

JeffreyWay:
@roddypratt Checking that discussion, it still seems like folks are having problems - even with 4.6.1. So I'm going to hold off a bit longer.

LeoColomb:
> So I'm going to hold off a bit longer.

Then is it possible to keep this issue open? Will make it easier to follow resolution for everyone.

> Checking that discussion, it still seems like folks are having problems

Which discussion? On https://github.com/sass/node-sass/pull/2147 nothing has been added/reported since release 4.6.1...

Thx! 👍

JeffreyWay:
https://github.com/sass/node-sass/issues/2139

bungcip:
I also got the same error when upgrading my node to 9.2 in windows 10. Is there is any workaround beside downgrading my node version?

[...]

PR #1339, opened 2017-11-23
- fix node-sass

hoquangthaiholy:


ankurk91:
Ref #1338 #1315 #1334

LeoColomb:
Ref #1306

PR #1340, opened 2017-11-25
vueExtractPlugin can use a function

Shenfq:
i think that vueExtractPlugin can use a function.
sometime,i have multiple entry js, i want export multiple css file, like this:
```
mix.options({
    extractVueStyles: '[name].css'

mix
.js('resources/assests/a/index.js', 'public/js/a.js')
.js('resources/assests/b/index.js', 'public/js/b.js');
```
This can generate two css files(a.css、b.css) in `public/js/` directory,but I want these two files are in another directory, like `public/css/`.

If i can use the function parameter,i can do that.
```
mix.options({
    extractVueStyles: function (getPath {
        return getPath('[name].css').replace('/js', '/css');
    }
}

mix
.js('resources/assests/a/index.js', 'public/js/a.js')
.js('resources/assests/b/index.js', 'public/js/b.js');
```
and `extract-text-webpack-plugin` can use function parameter
```
new ExtractTextPlugin(options: filename | object);
filename: {String\\|Function}
```

so, i pull request to solve this problem.

### DIFF
--- a/src/builder/webpack-rules.js
+++ b/src/builder/webpack-rules.js
@@ -202,9 +202,20 @@ module.exports = function () {
     let vueExtractPlugin;
 
     if (Config.extractVueStyles) {
-        let fileName = typeof(Config.extractVueStyles) === "string" ? Config.extractVueStyles : 'vue-styles.css';
-        let filePath = fileName.replace(Config.publicPath, '').replace(/^\//, "");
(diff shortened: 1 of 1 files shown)


laracasts:
Thanks so much for the PR. For now, I'd prefer to keep it simpler, since yours is the first request we've ever received for this. I'll keep in mind though. Thanks!

Issue #1226, opened 2017-09-29
Paths are resolved inconsistently (are they?)


lmeysel:
Having the following config
```
mix.setPublicPath(__dirname + '/output')
    .js('input-foo/app.js', 'js')
    .sass('input-foo/layout.scss', 'css');
```
mix generates the subfolder /output/css where the css files are stored, `/output/images` for the images referenced from the css, but the JS is bundled into `/output/output/js`.

In fact, the whole subfolder structure is recreated and then /js added - i.e.
```
mix.setPublicPath(__dirname + '/foo/bar/baz')
     .js('input-foo/app.js', 'js');
```
leads to the js-path `/foo/bar/baz/foo/bar/baz/js`. JS is bundled into the right folder, when i set the public path to the project root, but then the fonts- and images-folders are not created in the subdirectory i want it to be.

(I wonder how that works when using mix included in a laravel-application, because there the public path is not set explicitly, but the fonts- and images-folders are created in the public folder. When using mix in a non-laravel-environment mix crashes when not setting the public path explicitly).

I am running on windows 10 machine, might be interesting due to the fact we're talking about paths...

**EDIT:** After reading #1153 i tried to set public path not absolute but with windows' path-separator (i.e. backslash), now everything works as expected: setPublicPath('foo\bar\baz') - but it must not have a leading or trailing slash.

webmarkelov:
I have a problem, too.

lmeysel:
I also found a workaround/solution for that. The problem was the path-separator. When `mix.setPublicPath('output/assets')`, the problem is like described in my first post. But when `mix.setPublicPath('output'+path.sep+'assets')`, everything works fine for me (but not when using path.resolve(..) though).

grenaria:
This issue is not only for windows. I am having the same thing happen when the path for setPublicPath is an absolute path inside of a CentOS Vagrant box. Everything else compiles fine into the absolute public path (SCSS, Vue, etc) but the app.js from the js command gets put into a directory relative to the executable path with the absolute path appended to it.

Issue #1153, opened 2017-08-31
Issue with output paths when combining setPublicPath & setResourcePath


christopherdarling:
- Laravel Mix Version: 1.4.2
- Node Version (`node -v`): 8.4.0 
- NPM Version (`npm -v`): 5.2.0
- yarn Version: 0.27.5
- OS: Windows 10

### Description:
Output paths incorrect

### Steps To Reproduce:
*folder structure*
- /webpack.mix.js
- /themes/default/dev/js/main.js
- /themes/default/dev/scss/main.scss
- /themes/default/dev/svg/**.svg

I would like all dist files to be placed in themes/default/dist/ ideally in js, css, images, fonts subfolders.

*webpack.mix.js*
```js
mix.setPublicPath('./themes/default/dist/');
mix.setResourceRoot('/themes/default/dist/');

mix.js('./themes/default/dev/js/main.js', 'js');
mix.css('./themes/default/dev/scss/styles.scss', 'css');
```

*styles.scss*
contains 
```css
.class {
background-image: url('../svg/background-image.svg');
}
```

This results in the following output
- /themes/default/dist/css/styles.css
- /themes/default/dist/fonts/ (with the .svg)
- /themes/default/dist/themes/default/dist/js/main.js
- /themes/default/dist/manifest-mix.json - contents below
```js
{ 
    "/themes/default/dist/js/main.js": "/themes/default/dist/js/main.js", 
    "/css/styles.css": "/css/styles.css"
}
```

So the JS files are being output in the incorrect folder structure.

I have tried;
1. removing `setResourceRoot` - svgs still output in the same place but incorrect path in css file
2. setting public path to `./` and updating dist paths in `mix.js()` / `mix.css()` to include the full output path this results in the svg path in `styles.css` being `/themes/default/dist/fonts/background-image.svg` but the file located `/fonts/background-image.svg` - but the javascript now outputs to the correct path.

Happy to try out other configurations / run tests

christopherdarling:
I've managed to have made some progress on this, it looks as though there is a issue with Windows paths

*new webpack.mix.js*
```js
mix.setPublicPath('themes\\default\\dist');
mix.setResourceRoot('/themes/default/dist/');

mix.js('./themes/default/dev/js/main.js', 'js\\main.js');
mix.css('./themes/default/dev/scss/styles.scss', 'css\\styles.css');
```

This results in the correct file structure and css url() asset paths

PR #1343, opened 2017-11-27
Fix override the default path public directory on windows

webmarkelov:
https://github.com/JeffreyWay/laravel-mix/issues/1153
https://github.com/JeffreyWay/laravel-mix/issues/1226


Scilaris:
Nice solution! Works just fine on Win10 64

Issue #1345, opened 2017-11-29
Production sourcemaps not working with uglifyjs


stephan-v:
- Laravel Mix Version: 1.6.0 (`npm list --depth=0`)
- Node Version (`node -v`): 8.9.0
- NPM Version (`npm -v`): 5.0.3
- OS: Ubuntu 16.04

### Description:

Mix will not output a sourcemap for javascript files when running:

    npm run production

### Steps To Reproduce:

- Create a new laravel installation
- Update the webpack.mix.js file:

```
mix.js('resources/assets/js/app.js', 'public/js')
   .sass('resources/assets/sass/app.scss', 'public/css');

mix.webpackConfig({
    devtool: 'source-map'
});
```
Run:

    npm run prodution

It will break, Javascript sourcemaps will not be generated. Uglify also seems to strip the sourceMappingUrl comment from the original Javascript files.

As soon as you add:

```
mix.options({
    uglify: false
});
```
Everything seems to work fine. 

It seems as the devtool option from Webpack does not play nice with the uglifyjs-webpack-plugin. Since almost nobody uses source maps in production I am guessing this does not come up often.

ankurk91:
ref
#1344

PR #1346, opened 2017-11-29
Production sourcemaps fix

stephan-v:
It seems you also need to still add the `sourceMap: true` to the options.

https://github.com/mishoo/UglifyJS2/tree/harmony#source-map-options

The docs are not very clear about this because they want to be supplied with an Object but I have tried it with true and it seems to work.

If you don't set it, sourcemaps will fail in production:

https://github.com/JeffreyWay/laravel-mix/issues/1345

Try a clean Laravel installation and add:

```
mix.options({
    sourcemaps: 'source-map'
});
```

You will see that it will not create javascript sourcemaps when you run `npm run production` because it clashes with uglify.

This will work though:

```
mix.options({
    sourcemaps: 'source-map',
    uglify: {
        sourceMap: true,
        uglifyOptions: {
            sourceMap: true,
            compress: {
                warnings: false,
                drop_console: true,
            },
            output: {
                comments: false
            }
        }
    }
});
```

Please confirm this behavior, but it seems to fix the issue.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -276,6 +276,7 @@ module.exports = function () {
         uglify: {
             sourceMap: true,
             uglifyOptions: {
+                sourceMap: true,
                 compress: {
                     warnings: false,
                     drop_console: true,


JeffreyWay:
So we have to set `sourceMap: true` in two places? Surely that's not right, is it?

dp88:
The sourcemaps produced here still aren't working as expected for me. It maps the minified code to a version of the same file before whitespace was removed, rather than to a full list of all of my original source files.

```javascript
{"version":3,"file":"/assets/main.js","sources":["webpack:////assets/main.js"],"sourcesContent":["(minified sourcecode)"]}
```

In dev builds, the `sources` array contains an exhaustive list of all of my source files, and the map works properly. In prod builds, the map is now generated, but it does not actually contain information usable by the debugger.

stephan-v:
I will investigate it further, the mapping also does not seem entirely correct for me. The sourcemaps are properly fucked for production usage though, so something needs to change.

I will try to get a solid fix for this but I would appreciate any help I can get with this.

12 ways for Webpack to create sourcemaps with the `devtool` property is enough to start giving me headaches.

PR #1357, opened 2017-12-05
Add option to config mix based on the environment

jm-mwi:
I needed some options beyond whether mix was in production or not. So I added an option for mix to check it's environment against an environment name provided by the user.

So you can now do something like...
```js
if(mix.isEnvironment('development')) { ... }
```


JeffreyWay:
I think we'll pass on this for now. But you can always check `Config.env` in your webpack.mix.js file.

PR #1358, opened 2017-12-06
Allow for users to disable ExtractTextPlugins via an option

SethTompkins:
This allows for generated css files to be injected when using `npm run hot`.

To test: use a hot reloading setup, with a .sass() call. Change file and see injection as if you were using BrowserSync.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "laravel-mix",
-  "version": "1.6.1",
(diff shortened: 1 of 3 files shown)


JeffreyWay:
I'd rather we not require the user to know anything about this setting. It should automatically work. 

So maybe we can update the code to automatically check if the user is running `hot`, and then set whether to disable the extract plugin.

PR #1367, opened 2017-12-11
Distinguish between image SVG and font SVG

tpraxl:
Both, fonts and images may be .svg files.

Before this commit, laravel-mix would consider all *.svg files to be font files. When resolved (e.g. because it was referred to in a scss file), they would have been copied to the fonts directory.

Now, laravel-mix distinguishes between the two types by considering their names or paths:

* font recognition pattern: `font.*\.svg$` will resolve font.svg, font2.svg and font/awesome.svg
* image recognition pattern: `im[a]?g.*\.svg$` will resolve img/1.svg and img1.svg

Images will be resolved and copied into the `images` directory, fonts will be resolved and copied into the `fonts` directory.

### DIFF
--- a/src/builder/webpack-rules.js
+++ b/src/builder/webpack-rules.js
@@ -65,7 +65,7 @@ module.exports = function () {
 
     // Add support for loading images.
     rules.push({
-        test: /\.(png|jpe?g|gif)$/,
+        test:  /(\.(png|jpe?g|gif)$|im[a]?g.*\.svg$)/,
         loaders: [
             {
                 loader: 'file-loader',
@@ -95,7 +95,7 @@ module.exports = function () {
 
     // Add support for loading fonts.
     rules.push({
-        test: /\.(woff2?|ttf|eot|svg|otf)$/,
+        test: /(\.(woff2?|ttf|eot|otf)$|font.*\.svg$)/,
         loader: 'file-loader',
         options: {
             name: path => {
--- a/src/builder/webpack-rules.js
+++ b/src/builder/webpack-rules.js
@@ -65,7 +65,8 @@ module.exports = function () {
 
     // Add support for loading images.
     rules.push({
-        test:  /(\.(png|jpe?g|gif)$|im[a]?g.*\.svg$)/,
+        // only include svg that doesn't have font in the path or file name by using negative lookahead
+        test:  /(\.(png|jpe?g|gif)$|^((?!font).)*\.svg$)/,
         loaders: [
             {
                 loader: 'file-loader',
--- a/test/features/mix.js
+++ b/test/features/mix.js
@@ -191,6 +191,20 @@ test.cb.serial('the kitchen sink', t => {
     });
 });
 
+test.cb.serial('it resolves image- and font-urls and distinguishes between them even if we deal with svg', t => {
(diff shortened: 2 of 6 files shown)


kohenkatz:
I have been overriding the generated Webpack config to do this myself for a while, and I'm very happy to see a PR for it, but I see one issue - what happens to SVG files that aren't in a directory (or otherwise named with) with `font` or `im[a]?g`?

tpraxl:
> what happens to SVG files that aren't in a directory (or otherwise named with) with font or im[a]?g?

Thanks for the point. They wouldn't be picked up at all.  

Maybe I should update the PR for default handling of either fonts or images (I would vote for images).

To be honest: If it was for my opinion, I would remove the font-handling for svg files completely. But I followed some of the discussions and have perceived that Jeffrey Way wants to keep the support.

I would appreciate opinions on this: Should svg files that don't match any of these patterns be handled as images or as fonts?

kohenkatz:
I agree with leaving support for using SVG fonts - even though all modern browsers support WOFF and most of them support other options too - just because there's no reason not to.

That said, personally, I think anything that is not *explicitly* a font should be handled as an image.

tpraxl:
Thanks. I'll wait for some other options and will update the PR during the next days.

tpraxl:
@JeffreyWay Which option would you prefer? 

a) Consider an SVG an image unless its name has the font-pattern?
b) Consider an SVG a font unless its name has the image pattern?

tpraxl:
Just updated the PR to fix the following issue pointed out by @kohenkatz 

If an svg did not contain the words img, image or font in the path or filename, it would not have been picked up at all.

Per default, all svg files are now considered images, **except** when the word "font" appears in the path or filename.

So the following files would be considered fonts for example:

* font/awesome.svg
* fonts/awesome.svg
* assets/font.svg

All other svg files would be considered images.

I also extended the test to verify that the files are not accidently handled by both webpack rules.

The image rule is implemented using a negative lookahead.

JeffreyWay:
Okay, I'm going to merge this. 

Thanks for taking care of it, @tpraxl! I'll make a new release this week.

JeffreyWay:
@tpraxl So what do you think... is this enough of a breaking change that we should tag a major release? Not sure.

tpraxl:
Yes I think it justifies a major release.

tpraxl:
Projects that dealt with images ending up in the fonts folder might have used static references like ``. So it's a breaking change that should justify the major version. I would expect that lots of projects went this way instead of using `{{ mix(…) }}`, because I think that laravel-mix is used outside laravel a lot.

JeffreyWay:
Yeah good point. Okay, 2.0 it is.

aminraeisi:
@tpraxl I just updated today and saw this change in my changed files. I have some place holder images that are stored in `storage/app/public/photos/i/` and I use them when there is no image uploaded by the user (profile images/shops logo/user image/etc).
All the placeholders are called in my blade files except one that was in my vuejs file. Now I have only that image created in my `public/images/` folder.
So the src that I had before was:
``
But with this update I can do this:
``
Which is better but only helps with one of the placeholders. Is there a way to have those other images like this?

PR #1369, opened 2017-12-13
Remove unused asset parameter

sebastiandedeyne:
I assume this is a remainder of an old implementation, but the `asset` parameter is unused here.


JeffreyWay:
Yep, good catch. Thanks!

PR #1378, opened 2017-12-22
Vue loader updated

pavinthan:
Vue loader updated

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "style-loader": "^0.18.1",
     "uglify-js": "^2.8.28",
     "uglifyjs-webpack-plugin": "^1.0.0",
-    "vue-loader": "^13.0.5",
+    "vue-loader": "^13.6.1",
     "vue-template-compiler": "^2.0.0",
     "webpack": "^3.5.0",
     "webpack-chunk-hash": "^0.4.0",


pavinthan:
Thank you guys 👍

askaoru:
I'm not too sure which are at fault here but i'm getting the error and I think it's kinda related to this PR.
```
Vue packages version mismatch:

- vue@2.5.13
- vue-template-compiler@2.5.9

This may cause things to work incorrectly. Make sure to use the same version for both.
If you are using vue-loader@>=10.0, simply update vue-template-compiler.
If you are using vue-loader@<10.0 or vueify, re-installing vue-loader/vueify should bump vue-template-compiler to the latest.
```

I'm using laravel-mix 1.7.2

ruchern:
@askaoru Your issue is with `vue-template-compiler` and not `vue-loader`

Just run `npm install -D vue-template-compiler@latest` or yarn and you will be fine.

askaoru:
Thanks @ruchern .
But for some reason I need to include both
`"vue-template-compiler": "^2.5.13"` and `"vue-loader": "^13.6.1"`.

Just adding the latest `vue-template-compiler` gave me a `Module not found: Unable to resolve vue-loader`, which is pretty odd. I'll take a look later if there's any conflict my other libraries.

ruchern:
@askaoru That is indeed weird but your method does solve the problem.

PR #1392, opened 2018-01-05
Allow development in node_modules

### LINKED ISSUE #1379: Removing node_modules from babel exclude config
- Laravel Mix Version: 1.7.2
- Node Version: 8.1.4
- NPM Version: 5.4.2
- OS: Windows 10 (Git Bash Terminal)

### Description:

I'm currently working on a project where we are also developing a library alongside it using jsx. Here's the setup:

```jsx
// /system/resources/assets/js/Icones.jsx
import ReactDOM from 'react-dom';
import { Icon } from '@nitrodigital/guia-de-estilos'; //not npm install'ed, but npm linked to folder /guia-de-estilos

export default class Icones {
	init() {
		let icons = document.querySelectorAll('[data-icone]');
		icons.forEach(function(root) {
			let icon = root.dataset.name;
			ReactDOM.render(<Icon name={icon} />, root);
		});
	}
}

// /guia-de-estilos/src/Icon.jsx
import React from 'react';

export default class Icon extends React.Component {
	constructor(props) {
		super(props);

		this.icons = {
			user: (
				<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24">
					// a REALLY LONG path, not relevant
				</svg>
			)

		};
	}

	render() {
		return this.icons[this.props.name];
	}
}
```

Running `npm run dev` I got the following error, which is exactly at the starting `<svg>`.
```
ERROR in ./node_modules/@nitrodigital/guia-de-estilos/src/Audrye/Icon.jsx
Module parse failed: Unexpected token (9:4)
You may need an appropriate loader to handle this file type.
```
After some search, I got it working by modifying the `node_modules/laravel-mix/src/builder/webpack-rules.js` by removing the `node_modules` exclusion from the jsx compilation rules.

```js
// From
    // Babel Compilation.
    rules.push({
        test: /\.jsx?$/,
        exclude: /(node_modules|bower_components)/,
        use: [
            {
                loader: 'babel-loader',
                options: Config.babel()
            }
        ]
    });

// To
    // Babel Compilation.
    rules.push({
        test: /\.jsx?$/,
        exclude: /(bower_components)/,
        use: [
            {
                loader: 'babel-loader',
                options: Config.babel()
            }
        ]
    });
```

How do I remove that exclusion properly, without messing with laravel-mix files?

forsureitsme:
Fix #1379

### DIFF
--- a/src/builder/webpack-rules.js
+++ b/src/builder/webpack-rules.js
@@ -8,7 +8,7 @@ module.exports = function () {
     // Babel Compilation.
     rules.push({
         test: /\.jsx?$/,
-        exclude: /(node_modules|bower_components)/,
+        exclude: /(bower_components)/,
         use: [
             {
                 loader: 'babel-loader',


ankurk91:
It is package author duty to publish transpiled code so that end user no need to.

Some examples are:
https://unpkg.com/react@16.2.0/
https://unpkg.com/angular@1.6.8/
https://unpkg.com/vue@2.5.13/
Notice that their `package.json` **main** property points to a transpiled js file.

Allowing babel-loader to go through node-modules folder will increase build time.
You can override mix configs with `mix.webpackConfig`
https://github.com/JeffreyWay/laravel-mix/blob/master/docs/quick-webpack-configuration.md

JeffreyWay:
Exactly what @ankurk91 said. It would impact performance significantly.

Issue #1373, opened 2017-12-16
Config.publicPath cannot contain dot characters


unclecheese:
- Laravel Mix Version: 1.7.2
- Node Version 8.9.2
- NPM Version 5.5.1
- OS: macOS 10.13

### Description:

If `Config.publicPath` contains a dot character, e.g. `setPublicPath('./')`, this ends up getting parsed as unescaped regex in `Manifest.noramlizePath()`, which ends up truncating the first character of the path.

### Steps To Reproduce:

_webpack.mix.js_
```
mix.setPublicPath('./')
  .scss('path/to/main.scss', 'path/to/dist.css');
```

Resulting `manifest.json`:
```
{
  '/ath/to/dist.css': '/ath/to/dist/css'
}
```

Granted, setting `./` as a public path is probably redundant in most cases (I removed it and it works fine), but it seems to me the `mix` API should be robust enough that it doesn't lead to bizarre results like this that leave devs digging through internals to figure out what configuration error they made.

gummi-io:
I have the same issue, having dot in the public path works before 1.4.2 tho. Not sure which version does it stop working

JeffreyWay:
Fixed.

Issue #1326, opened 2017-11-18
mix.version() error: Error: ENOENT: no such file or directory, open '/css/app.css'


sebastiansulinski:
I'm using standalone version of `laravel-mix@1.6.1` with `npm@5.5.1` and I'm only compiling `scss` file:

```javascript
let mix = require('laravel-mix');

mix
    .setPublicPath('/')
    .sass('scss/app.scss', 'css/')
    .sourceMaps()
    .version();
```

When run with `version()` it throws an error:

```
fs.js:646
  return binding.open(pathModule._makeLong(path), stringToFlags(flags), mode);
                 ^

Error: ENOENT: no such file or directory, open '/css/app.css'
    at Object.fs.openSync (fs.js:646:18)
    at Object.fs.readFileSync (fs.js:551:33)
...
```

When I remove call to `version()` - it does compile without any issues, but of course, manifest does not have version parameter.

I also tried to changing the public path to the relative `./`:

```javascript
let mix = require('laravel-mix');

mix
    .setPublicPath('./')
    .sass('scss/app.scss', 'css/')
    .sourceMaps()
    .version();
```
but this has resulted with the same error (except for the path, which now was showing as absolute) and `mix-manifest.json` has cut first letter from the directory - still without the version present:

```json
{
    "/ss/app.css": "/ss/app.css"
}
```

sebastiansulinski:
Changing to `standaloneSass('scss/app.scss', 'css/')` magically resolved the issue.

```javascript
mix
    .setPublicPath('./')
    .standaloneSass('scss/app.scss', 'css/')
    .sourceMaps()
    .version();
```

sebastiansulinski:
It seems I've closed it too early.
After compiling for production, I've noticed that it messed up some styling - adding :hover event to many elements - looks like `standaloneSass` might be having some trouble parsing some scss.

When I tried to compile for production with `sass` but without `version` - everything works fine except for the fact that I cannot have versioning in manifest, which is quite important for my project.

sinchang:
```
let mix = require('laravel-mix')

mix.js('src/js/app.js', 'public/js')
  .sass('src/sass/app.scss', 'public/css')
  .version()
```
can't work with this config, may I have the same problem.

`Error: ENOENT: no such file or directory, open '/public/js/app.js'`

bretterer:
Any update on this @JeffreyWay .. I am also having this problem... My config is 

```
mix.setPublicPath(`./`);

[...]

PR #1407, opened 2018-01-10
Editing Manifest normalizePath method to escape Config.publicPath special characters

DiaaEddin:
when I tried to set public path to root, every thing was okay. but I got error on version when running:
`npm run prod`

**Steps To Reproduce:**

webpack.mix.js
```
mix.setPublicPath('./');
mix.sass("Resources/sass/uikit/theme.scss", "content/css/uikit.css")
    .version();
```

Resulting manifest.json:
```
{
    "/ontent/css/uikit.css": "/ontent/css/uikit.css?id=ff7f39e6c3f32433eaee"
}
```
npm run prod
```
fs.js:642
  return binding.open(pathModule._makeLong(path), stringToFlags(flags), mode);
                 ^

Error: ENOENT: no such file or directory, open 'G:\git_projects\mobile-shop\MobileShop\ontent\css\uikit.css'

```

**Related issues** [#1373](https://github.com/JeffreyWay/laravel-mix/issues/1373) [1326](https://github.com/JeffreyWay/laravel-mix/issues/1326)

but after debugging the code the problem was in manifest.js normalizePath and escaping special characters in Config.publicPath magically solved the issue.

### DIFF
--- a/src/Manifest.js
+++ b/src/Manifest.js
@@ -121,9 +121,10 @@ class Manifest {
      * @param {string} filePath
      */
     normalizePath(filePath) {
-        filePath = filePath.replace(
-            new RegExp('^' +  Config.publicPath), ''
-        ).replace(/\\/g, '/');
+        if (Config.publicPath && filePath.startsWith(Config.publicPath)) {
+            filePath = filePath.substring(Config.publicPath.length);
+        }
+        filePath = filePath.replace(/\\/g, '/');
 
         if (! filePath.startsWith('/')) {
             filePath = '/' + filePath;


andyblackwell:
Maybe it would be better to make it simpler, rather than more complex? Remove the regex entirely.
```
if (filePath.startsWith(Config.publicPath))) {
    filePath = filePath.substring(Config.publicPath.length);
}
filePath = filePath.replace(/\\/g, '/');
```

JeffreyWay:
Yeah let's go with @andyblackwell's suggestion.

DiaaEddin:
@andyblackwell thank you for your suggestion, I've tested your edit, and committed it as it is. 
my code added more complexity because I thought it is better to keep the changes to minimum.

Issue #1420, opened 2018-01-19
When attempting to compile any more than 9 .less files, the number of resulting .css files are always one short


susanBuck:
- Laravel Mix Version: 1.7.2
- Node Version: 9.4.0
- NPM Version: 5.6.0
- OS: MacOS High Sierra

### Description:
**When attempting to compile any more than 9 less files (via `mix.less`), the number of resulting .css files are always one short.**

I ran the following tests:
+ If I try to compile 12 .less files, only 11 css files are produced
+ If I try to compile 11 .less files, only 10 css files are produced
+ If I try to compile 10 .less files, only 9 css files are produced
+ BUT If I try to compile 9 .less files, then the correct amount of css files are produced, 9.
+ Likewise, if I compile 8 .less files, then the correct amount of css files are produced, 8.

### Steps To Reproduce:
I replicated my problem in a fresh Laravel install which is available here:
<https://github.com/susanBuck/foobar>.

`webpack.mix.js` is as follows:
```js
let mix = require('laravel-mix');

mix.less('resources/less/a.less', 'css/')
    .less('resources/less/b.less', 'css/')
    .less('resources/less/c.less', 'css/')
    .less('resources/less/d.less', 'css/')

    .less('resources/less/e.less', 'css/')
    .less('resources/less/f.less', 'css/')
    .less('resources/less/g.less', 'css/')
    .less('resources/less/h.less', 'css/')

    .less('resources/less/i.less', 'css/')
    .less('resources/less/j.less', 'css/')
    .less('resources/less/k.less', 'css/')
    .less('resources/less/l.less', 'css/');
```

And below are the details of 5 different tests that vary the amount of `mix.less` calls by commenting them out.

Between each test, I run this command...

```bash
rm -rf ./public/css; npm run dev; find ./public/css -type f -and -not -name ".DS_STORE" | wc -l
````

...in order to:

1. Wipe `public/css` (destination folder) clean for a blank slate
2. Run `npm run dev` to invoke Mix
3. Report the number of resulting .css files that were produced in `public/css`

**Test 1:**
Leave all 12 .less statements as is.
Invoke above-mentioned command.
.css file count:
+ Expected: 12
+ Actual: 11
_FAIL_

**Test 2**
Comment any ONE of the .less statements so only 11 are run.
Invoke above-mentioned command.
.css file count:
+ Expected: 11
+ Actual: 10
_FAIL_

**Test 3**
[...]

PR #1424, opened 2018-01-19
Find approach sometimes deletes the wrong file

susanBuck:
Discovered a weird bug where compiling > 9 .less files was always produced .css file short.
Traced it down to this function where it was deleting a compiled .css file instead of the intended `mix.js`.
Full details here: https://github.com/JeffreyWay/laravel-mix/issues/1420


JeffreyWay:
Sorry, Susan. I pushed a fix up just a few minutes ago without seeing this. Thanks for finding the issue, though. Much appreciated.

susanBuck:
No problem, glad it's sorted out.

PR #1425, opened 2018-01-20
Feature/automated testing for pull requests

tpraxl:
# Allow for automated testing of pull requests

The tests are not only run on **Linux** + **OSX**, but also on **Windows**, with the help of AppVeyor. 

* This helps to prevent laravel-mix from breaking on one of these systems (given the coverage is sufficient)
* forces contributors to support the other platforms 
* and creates confidence in PR changes

The PR not only contains the configurations for [Travis CI](https://travis-ci.org/) and [AppVeyor](https://www.appveyor.com/) but also **fixes**

* old test failures on windows
* randomly failing tests
* and one failing test, that was introduced in one of the latest merges (see commit history of this PR for details).

Note that the `test` script has been changed in `package.json` for two reasons:

* `sudo` isn't required to run nyc or ava
* `--serial` is used to prevent filesystem race conditions and thus randomly failing tests (Test A creates a file, Test B picks it up, Test A deletes it while Test B reads it)

The commit messages often contain details to explain why a certain change was made.

In order to automate the tests, you need to configure the github project: 

## Travis CI

`Settings -> Integrations & Services -> Add service -> Travis CI`

## AppVeyor

You need to create an account. Like Travis CI, this is free for open source projects. 

You then need to add your project via `+ New Project -> GitHub -> laravel-mix`. In the process, AppVeyor will setup the webhook (`Settings->Webhooks: https://ci.appveyor.com/api/github/webhook (pull_request and push)`) in GitHub for you.

## Example reports

* Travis CI: https://travis-ci.org/tpraxl/laravel-mix/builds/331173523
* AppVeyor: https://ci.appveyor.com/project/tpraxl/laravel-mix/build/1.0.32

Thanks to @kohenkatz, who helped a lot with fixing issues and failing tests on windows (see #1383).

Notice that [OSX builds currently have very long delays on Travis CI](https://www.traviscistatus.com/). It took almost 7 hours until all the OSX instances have started for the last commit. This should be a temporary issue though.

PR #1426, opened 2018-01-21
Update packages to fix security vulnerabilites

ruchern:
screen shot 2018-01-21 at 12 58 44 pm

The screenshot of the problem.


tpraxl:
This merge broke the tests. See [Travis CI](https://travis-ci.org/tpraxl/laravel-mix/jobs/331456438) and [AppVeyor](https://ci.appveyor.com/project/tpraxl/laravel-mix/build/1.0.33). This issue is also a good reason for merging #1425 IMHO.

tpraxl:
Also, why has the yarn.lock been removed?

ruchern:
On breaking the tests, I did not change anything on the test. I checked the test before and after and it's the exact same results. I think the test is already broken before I did anything. I would love to fix it, but I am very new to testing for now.

On the topic of `yarn.lock` being removed, I was the one who merged the `yarn.lock` by accident into the repo because I am using yarn. I do think we will need that and lets keep it simple to one "package lock" file.

tpraxl:
@ruchern my comment was no offense, really. The tests include comparison of file hashes and the file hash changed because of the library updates. This is normal. You didn't break laravel-mix, only the tests failed. I fixed that in #1425 to make sure that the automatic tests work immediately, should @JeffreyWay merge my PR. It's just a minor matter.

I also restored the yarn.lock in that PR, because it's important for deterministic test data currently (especially in node 6 environment on windows for AppVeyor).

ruchern:
Right, it makes sense that the file hash changes when the libraries are updated. Should that be more dynamic in this case since updating packages will change the hashes regularly and subsequently failing the test.

tpraxl:
Yes, I know that this test will fail over and over again after library updates. And I'm not sure what the intention for this test was originally (maybe I should watch the respective episode on laracasts again :) ). Just wanted to add automatic testing for all supported platforms and versions with reduced impact on the rest of the code.

I believe that – once automated tests are integrated – there will be discussions about tests that may not be 100% optimal and finally the situation will improve (I will gladly contribute to that). However, I didn't address this very test in this PR, because that's another story that should be discussed elsewhere to avoid blurring the focus of this PR.

PR #1436, opened 2018-01-25
can't change sourceMap type when production

Shenfq:
Sometimes, I use `mix.sourceMaps(true, 'hidden-source-map');` can't useful.

when production, I can't customize the type of sourceMap.

I commit the PR，want to solve this problem.

### DIFF
--- a/src/Api.js
+++ b/src/Api.js
@@ -375,9 +375,15 @@ class Api {
      * @param {Boolean} productionToo
      * @param {string}  type
      */
-    sourceMaps(productionToo = true, type = 'eval-source-map') {
+    sourceMaps(productionToo = true, type) {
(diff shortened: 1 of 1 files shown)


stephan-v:
How is this still open when this was submitted Jan 25? This should either be merged or closed...

JeffreyWay:
@Shenfq Thanks for your patience. This PR wasn't quite right, and broke our sourcemap tests. I've added an optional third argument to mix.sourceMaps() for the desired production sourcemap type to use. It defaults to 'source-map'.

```js
// argument order: productionToo, devType, productionType

mix.sourceMaps(true, 'eval-source-map', 'hidden-source-map');
```

Issue #1452, opened 2018-02-03
Custom mix manifest file


VinceG:
- Laravel Mix Version: 2.0.0
- Node Version (`node -v`): 8.7.0
- NPM Version (`npm -v`): 5.6.0
- OS: Mac OS X

### Description:

Seems like this feature was dropped in the rewrite, what this does is basically allows to specify multiple mix files to process and the mix-manifest.json file will update rather than re-created from scratch every time.
This allows to have separate applications on the same installation. 

### Steps To Reproduce:

This repository will demonstrate the issue and the fix applied:

https://github.com/VinceG/laravel-mix-issue-new

The following pull request was approved and merged but somehow the feature got removed during the rewrite since the 0.12 release.

https://github.com/JeffreyWay/laravel-mix/pull/609

I've added this issue as a reference a pull request will follow to the 2.0.0 branch to bring this feature back.

onnsoft:
I found solution for this. It's look like:

I have 2 files:

webpack.frontend.mix.js

```
const { mix } = require('laravel-mix');

mix.setPublicPath('public/frontend');
...
```
webpack.backend.mix.js

```
const { mix } = require('laravel-mix');

mix.setPublicPath('public/backend');
...
```

Separate run webpack for building: backend and frontend:

package.json
```
{
 ...
  "scripts": {
    "frontend-dev": "npm run frontend-development",
    "frontend-development": "cross-env NODE_ENV=development node_modules/webpack/bin/webpack.js --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js --env.mixfile=webpack.frontend.mix.js",
    "frontend-watch": "cross-env NODE_ENV=development node_modules/webpack/bin/webpack.js --watch --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js --env.mixfile=webpack.frontend.mix.js",
    "frontend-watch-poll": "npm run watch -- --watch-poll",
    "frontend-hot": "cross-env NODE_ENV=development node_modules/webpack-dev-server/bin/webpack-dev-server.js --inline --hot --config=node_modules/laravel-mix/setup/webpack.config.js --env.mixfile=webpack.frontend.mix.js",
    "frontend-prod": "npm run frontend-production",
[...]

PR #609, opened 2017-03-23
Added the ability to specify a custom webpack.mix.js file as well as …

VinceG:
This addresses the following two issues:

#607 
#608

I can comment to describe why i've made the changes the way they are. it might seem like it doesn't make sense but it's been working.

This has been working and tested on a public repository you can test this branch with

https://github.com/VinceG/laravel-mix-issue-manifest

PR #1453, opened 2018-02-03
Laravel mix multi

VinceG:
Details:
https://github.com/JeffreyWay/laravel-mix/issues/1452

This is a simple addition that will allow to run multiple applications on the same setup. For example we use laravel and laravel mix. we have multiple *.mix.js files that build different parts of the website, each time we build it it updates the mix-manifest.json file, however the feature was removed from the time it was merged at v.0.12.0 till now.

The changes below will bring this back so we can have multiple mix files and build multiple applications on the same installation. it basically _merges_ the mix-manifest.json file vs. _overwriting_ it from scratch at every build. This prevents every build from deleting previous builds.

As an example:

**Before:**

```
NODE_ENV=development node_modules/webpack/bin/webpack.js --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js "--env.mixfile=build/first.mix"
```
mix-manifest.json looks like this:

```
{
    "/first.js": "/first.js",
    "/first.css": "/first.css"
}
```

```
NODE_ENV=development node_modules/webpack/bin/webpack.js --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js "--env.mixfile=build/second.mix"
```

mix-manifest.json looks like this:

```
{
    "/second.js": "/second.js",
    "/second.css": "/second.css"
}
```

**After**

```
NODE_ENV=development node_modules/webpack/bin/webpack.js --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js "--env.mixfile=build/first.mix"
```
mix-manifest.json looks like this:

```
{
    "/first.js": "/first.js",
    "/first.css": "/first.css"
}
```

```
NODE_ENV=development node_modules/webpack/bin/webpack.js --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js "--env.mixfile=build/second.mix"
```

mix-manifest.json looks like this:

```
{
    "/first.js": "/first.js",
    "/first.css": "/first.css",
    "/second.js": "/second.js",
    "/second.css": "/second.css"
}
```

Notice the mix-manifest.json kept the first elements and appended the second to it.


MohannadNaj:
Love it. except I think it should be a configured option, and not by default.

a major issue of this PR: the existence of files at the manifest while it's not really exist. If at sometime during development I did this in `webpack.mix.js` :
``` js
    mix.js('resources/assets/js/experimental.js', 'public/js')
       .sass('resources/assets/sass/experimental.scss', 'public/css');
```

the manifest now will have:
``` js
{
    "/js/experimental.js": "/js/experimental.js",
    "/css/experimental.css": "/css/experimental.css"
}
```
then I changed that in my `webpack.mix.js` :
``` js
    mix.js('resources/assets/js/not-experimental.js', 'public/js')
       .sass('resources/assets/sass/not-experimental.scss', 'public/css');
```
the resulted manifest will be:
``` js
{
    "/js/experimental.js": "/js/experimental.js",
    "/css/experimental.css": "/css/experimental.css",
    "/js/not-experimental.js": "/js/not-experimental.js",
    "/css/not-experimental.css": "/css/not-experimental.css"
}
```
it will accumulate forever. correct me if I'm wrong.

However, I love this PR for two reasons:

1- It will solve your described issue.

2- It will solve rewriting the manifest every time after running the unit tests.
 
before running the unit test, I have this cute manifest:
``` js
 {
    "/js/main.js": "/js/main.js",
    "/css/main.css": "/css/main.css"
 }
```
after running the test, it will be:
``` js
{
    "/specs/setup/index.js": "/specs/setup/index.js",
    "/specs/my-test.spec.js": "/specs/my-test.spec.js"
}
```
Your PR will just merge them so I wouldn't be distracted about the changes 👍 

P.S : Looks like the failing tests is telling us a valuable info, you can't use `this.read()` directly without checking the existence of the manifest first.

tpraxl:
Merging makes the contents of the manifest file non-deterministic. And like @MohannadNaj said, it will accumulate forever and requires you to manually delete it for a production build.

A better solution would be to add scopes to the manifest file, like so: 

```
{
   first: {
      "/first.js": "/first.js",
      "/first.css": "/first.css"
   },
   second: {
      "/second.js": "/second.js",
      "/second.css": "/second.css"
   }
}
```

IMHO, an even better solution would be configurable manifest file locations and/or names for different builds. 

* first/first-mix-manifest.json
* second/second-mix-manifest.json

My proposals potentially require the laravel mix php helper function to be updated to allow for multi builds.

I generally appreciate your thoughts on multi builds. This might come in handy when dealing with service-workers. However, I would prefer a more deterministic approach.

VinceG:
@MohannadNaj Yes, that will be an issue, this is something that can be fixed but will be tricky and hacky way to accomplish something that the majority of users will not experience. 

For instance, we use this heavily, we have multiple sites (more than 10) with different build files JS/CSS etc, our mix manifest has 89 entries for ALL those websites and applications, We've been running this for over two years and to be honest the mix manifest file is never going to be thousands or even hundreds of lines to make it worth while going through each entry when building the manifest file and checking if the file exists or other type of hacky way to accomplish this.

I wasn't even aware of the tests, i can fix the failing tests and hopefully this can get merged.

@tpraxl This PR is just to be able to bring back what was already merged in earlier versions in the following PR #609, somewhere down the line of the Laravel Mix rewrite the feature was removed, this will just bring back what was already in there.

All these changes suggested to alter the mix-manifest will require changes to the Laravel framework code and the mix helper and this is not what this PR is about. Just like we did, there is always a way to further customize it and have your own version of it, for me personally we don't need the deterministic approach, we don't change or update the mix-manifest file that often or change the file names to have "orphaned" entries in the mix manifest file, For that reason i didn't even bother in doing any additional work. For the majority of developers this will be more than enough. 

As i mentioned before, this PR just brings back what we already had with Laravel Mix in earlier versions that was removed due to the rewrite (#609) Any additional work i think should be done separately as it goes beyond the scope of this PR.

*P.S Sorry for the long post*

VinceG:
@JeffreyWay Hey Jeffery, Any idea when we can have this looked at and merged?

Thank You.

ijpatricio:
hey guys, I'm also pursuing same goal, as it seems.

context is in a multi page app having different webpack pipelines, considering different node_modules folder, that are subdirs of the project root.

Slightly different, but is what it's beeing seeked here, right?

I already have a setup doing such, leveraging that php's mix helper already takes a 2nd argument for the manifest location, would this be interesting to see?

What i'm missing to achieve is to build all fronted end applications (several mix pipelines, all in one go)

PS: just for having different apps, multiple manifests will be ok?, just a matter of setting your desired public path

image

ijpatricio:
@MohannadNaj  @VinceG  @JeffreyWay 

just setup an example...

https://github.com/ijpatricio/laravel-mix-branches

it's still very manual, we have to just leave uncommented the one we want to build, in root webpack.mix.js

inspiring on https://www.npmjs.com/package/laravel-multimix

one could choose the 'area'...

npm run mix <module> <dev/prod>

what do you guys think?

VinceG:
@JeffreyWay  Any update on merging this PR?

Issue #1423, opened 2018-01-19
Overriding file loader rules to add exceptions


tegola:
I'm using [Quill](https://quilljs.com) in a Vue component. To avoid loading everything from the distributable build, I have prepared a quill entry file (like [this one](https://quilljs.com/guides/adding-quill-to-your-build-pipeline/#entry)).

When doing this, however, webpack moves the svg icons used by Quill theme in `/fonts/vendor/quill` and returns their path, instead of returning the svg content, as Quill expects. I end up with this:

![schermata 2018-01-19 alle 11 19 18](https://user-images.githubusercontent.com/1868590/35146259-c08d3a0e-fd0a-11e7-8358-476d765bf2af.png)

This doesn't happen when using the distributable build in the nom package.

Quill's [build guide](https://quilljs.com/guides/adding-quill-to-your-build-pipeline/) and [related webpack example](https://github.com/quilljs/webpack-example/) say that icons should be loaded with html loader:
https://github.com/quilljs/webpack-example/blob/master/webpack.conf.js#L39

**Now to my question:** how can I replace/override Laravel Mix's svg loader so it doesn't match Quill icon paths, and instead setup `html-loader` just for that path?

Thanks

**Edit:** A past issue (closed unanswered) asks the same thing: #942.

ankurk91:
Here is how you can add your own loaders.
```js
mix.webpackConfig({
module: {
    rules: [
     // your rules may go here
    // next rules are just example, you can modify them according to your needs
      {
        test: /\.(woff|woff2|eot|ttf|svg)(\?.*$|$)/,
        loader: 'file-loader?name=[name].[ext]?[hash]'
      }
    ]
 }
});
```
Not tested.

tegola:
That isn't going to work. I want to exclude the `.svg` files from being loaded by the default `fonts` rule in Laravel Mix – which is the one that handles SVGs – and add mine that uses a different loader.

If I just add my own rule for handling only SVGs (or SVGs within Quill's specific path), the default one executes first, and returns back the file path instead of its content.

JeffreyWay:
This should be fixed in the next release, which will be out today.

tegola:
Release 2.0 doesn't fix my issue, it just changes the output of quill SVGs in the images dir instead of the fonts dir.

tegola:
[...]

Issue #1397, opened 2018-01-07
CKEditor 5 SVGs do not build properly


ryanvade:
- Laravel Mix Version: 1.7.2
- Node Version: v8.9.4 
- NPM Version: 5.6.0
- Yarn Version: 1.3.2 
- OS: Fedora release 27

### Description:
I am attempting to use the new CKEditor 5 Framework with Mix.  As described in the [CKEditor Quick Start Guide](https://docs.ckeditor.com/ckeditor5/latest/framework/guides/quick-start.html) I have imported CKEditor libs into app.js (and removed everything else) then I built the file.  The results seem to show the SVGs are not loaded properly and the icons in the editor toolbar are missing. 

https://screenshots.firefox.com/LEYu0SnGQVMaew0E/127.0.0.1

I also receive this error output in the console.
```
XML Parsing Error: not well-formed
Location: http://127.0.0.1:8000/
Line Number 1, Column 1:
127.0.0.1:8000:1:1
```

I then attempted to fix the svg loader issue with the webpack config recommended on the guide.
```
mix.js('resources/assets/js/app.js', 'public/js').webpackConfig({
  module: {
    rules: [{
      test: /ckeditor5-[^/]+\/theme\/icons\/[^/]+\.svg$/,

      use: ['raw-loader']

    }]
  }
});
```
It produces the same result. I also tried the full webpack config:
```
mix.js('resources/assets/js/app.js', 'public/js').webpackConfig({
  module: {
    rules: [{
      test: /ckeditor5-[^/]+\/theme\/icons\/[^/]+\.svg$/,

      use: ['raw-loader']

    }, {
      test: /ckeditor5-[^/]+\/theme\/[^/]+\.scss$/,

      use: [
        'style-loader',
        'css-loader',
        'sass-loader'
      ]
    }]
  }
});
```
But I get errors:
```
ERROR  Failed to compile with 1 errors                                                                                                                                                                    17:29:16

 error  in ./node_modules/@ckeditor/ckeditor5-editor-classic/theme/theme.scss

Module build failed: 
// Copyright (c) 2003-2017, CKSource - Frederico Knabben. All rights reserved.
^
      Invalid CSS after "...load the styles": expected 1 selector or at-rule, was "var content = requi"
      in /home/ryan/Sites/ckeditortesting/node_modules/@ckeditor/ckeditor5-editor-classic/theme/theme.scss (line 1, column 1)

[...]

Issue #997, opened 2017-07-13
How to override file-loader rule?


Rigorbb:
How to exclude svg format from file-loader rule?

I tried to override it in mix.webpackConfig but it is not working

mix version is 1.2.0

JeffreyWay:
You can't at the moment.

Issue #949, opened 2017-06-29
overriding loaders


VinceG:
### Description:

Right now there is no way to override a load

if i set

```
mix.webpackConfig({
module: {
        rules: [
            {
                test: /\.jsx?$/,
                exclude: /node_modules\/(?!(vue-strap)\/).*/,
                loader: 'babel-loader' + mix.config.babelConfig()
            }
        ]
    }
});
```
nothing will happen and my loader will not get executed since they go by order and the original one defined in mix config will be used.

domstubbs:
It took forever to pin it down, but I eventually managed to get ES5 transpiling working with Foundation with the following config:


```js
mix.js('src/app.js', 'public/')
    .webpackConfig({
        module: {
            rules: [
                {
                    test: /\.jsx?$/,
                    exclude: /node_modules(?!\/foundation-sites)|bower_components/,
                    use: [
                        {
                            loader: 'babel-loader',
                            options: Config.babel()
                        }
                    ]
                }
            ]
        }
    });
```

hojjabr:
I'm getting this error after doing what @domstubbs wrote:

`webpack.mix.js:13 options: Config.babel()
                                     ^
ReferenceError: Config is not defined`

robbyrice:
Would you mind posting the contents of your webpack.mix.js file?

mediabeastnz:
Do we still need to do this? seems break without it on 6.4.3

mikerockett:
Same thing here…

```js
let mix = require('laravel-mix');

mix.webpackConfig({
    module: {
        rules: [{
            test: /\.jsx?$/,
            exclude: /node_modules(?!\/@<redacted>)/,
            use: [{
                loader: 'babel-loader',
                options: Config.babel()
            }]
        }]
    }
})
```

Config seems to be passed in via cli arg:

```
cross-env NODE_ENV=development node_modules/webpack/bin/webpack.js
    --watch --progress --hide-modules
    --config=node_modules/laravel-mix/setup/webpack.config.js
```

Unless that’s something else? Seems to compile fine on a Mac, but I’m on Windows...

tpraxl:
This Config seems to refer to the global Config object. Try using lowercase config, as this is exported by mix:

```js
// node_modules/laravel-mix/src/index.js
…
module.exports = api;
module.exports.mix = api; // Deprecated.
module.exports.config = Config;
```

PR #1466, opened 2018-02-10
Feature/override and merge rules

tpraxl:
Some use cases require modifications of existing webpack rules. It's no problem to add rules, but you can currently not reliably modify existing rules (except for hacks that might break in the future).

Related issues:

* #1423 (in my opinion, the most important one, because quill shouldn't be in the 20% use cases not supported by laravel-mix)
* #1397 (also an important one: CKEditor)
* #997 
* #949
* #942
* #374

This PR allows the user to replace, merge and skip the existing rules like this:

```js
// Skip rule
Mix.customizeRule.fonts = (rule, Config) => ({})
```

```js
// Replace rule
Mix.customizeRule.images = (rule, Config) => ({
    // completely replaces the original rule
    test: /\.png$/,
    loaders: [
      {
        loader: 'html-loader',
        options: {
          minimize: true
        }
      }
    ]
});
```

```js
// merge with own rules
Mix.customizeRule.images = (rule, Config) => {
    // only replace the test
    rule.test = /\.png$/;
    // add an exclude configuration
    rule.exclude = /node_modules\/quill\/assets\/icons\/(.*)\.svg$/;
    // and add a loader
    rule.loaders.push(myLoader);
    // the remaining original rule's configuration stays the same 
    return rule;
};
```

The "SVG images in font-folder" discussions would have been pointless if we had had this feature before.

Part of this PR is the restructuring of rules into their own respective files. See `src/rules` for all webpack rules used by laravel-mix (except dynamic preprocessor rules).

The user can hook into the rules by referring to their (file)names: `Mix.customizeRule[theRuleName]` and providing a closure for it. Mix will look up any defined closures and replace the prepared rule with the user defined rule.

I think this feature is important and we could very well raise "[80% of the use cases](https://github.com/JeffreyWay/laravel-mix/blob/master/docs/basic-example.md)" to 90% after merging this PR. 

So, I'm very motivated to continue working on it until possible doubts are resolved. 

Don't hesitate to ask for refinement.

[edit] corrected minor typo
[edit] added CKEditor issue


tpraxl:
This PR needs to be updated if/when #1467 gets merged.

tegola:
The `Mix.customizeRule.rulename = () => {}` looks a bit out of place in Mix. A more familiar syntax would look better and easier to understand. Something like `Mix.customizeRule('ruleName', (rule, Config) => {})`.

Also, new users reach out for Laravel Mix because webpack initial setup looks very confusing, and having them learn that `Mix` (the class) and `mix` (the instance in the docs) are different entities isn't ideal. Does `customizeRule` be used on the instance?

Again, great work.

ZapevalovAnton:
Yes, ```Mix``` and ```mix``` are misleading. Moreover ```Mix``` is not imported explicitly in the ```webpack.mix.js``` file. Maybe it is possible to use the version of @tegola ?

tpraxl:
Thanks for your feedback. I'll take your proposals into account (weird syntax and `mix` instead of `Mix`) and will update this PR asap.

tpraxl:
Sorry, didn't have the time to come back to this PR yet. It should still be considered active though.

tpraxl:
Added the option to use the syntax, suggested by @tegola and @ZapevalovAnton:

```js
mix.customizeRule('images', (rule, Config) => {});
```

tpraxl:
I reckon this PR is not going to be merged, having watched [this screencast](https://laracasts.com/series/whatcha-working-on/episodes/27). From what I've got so far, the idea of splitting it up into components is far better than just overriding rules – given the components can be dynamically added and their rules overridden. It fits into an idea I had while I prepared this PR: "recipes" for mix / webpack aspects.

tegola:
I'm not a subscriber, I can't see that screencast, sorry.

tpraxl:
@tegola see [the comparison between the branch he's working on with master](https://github.com/JeffreyWay/laravel-mix/compare/master...components)

tegola:
Aaaah bummer. Let's see how it turns out. Thanks.

tpraxl:
This PR conflicts with the plugin concept of laravel that has been developed in parallel or slightly later. I'll close it.

PayteR:
@tpraxl is there any documentation, how to substitute default components with my custom components? https://github.com/JeffreyWay/laravel-mix/issues/1616

raphaelsaunier:
I think an alternative, probably more convenient solution would be to use [webpack-chain](https://github.com/mozilla-neutrino/webpack-chain) (have a look at the [vue-cli docs](https://github.com/vuejs/vue-cli/blob/dev/docs/webpack.md) to see how it can be used) but I realise it would require some significant changes to the overall architecture of Laravel Mix.

---

Either way, if it helps anyone, here's how I ended up overriding the generated rules to correctly compile [quill.js](https://quilljs.com/guides/adding-quill-to-your-build-pipeline/), which didn't work out of the box with Mix:

```js
mix.extend('quill', webpackConfig => {
    const { rules } = webpackConfig.module;

    // 1. Exclude quill's SVG icons from existing rules
    rules.filter(rule => /svg/.test(rule.test.toString()))
        .forEach(rule => rule.exclude = /quill/);

    // 2. Instead, use html-loader for quill's SVG icons
    rules.unshift({
        test: /\.svg$/,
        include: [path.resolve('./node_modules/quill/assets')],
        loaders: [{ loader: 'html-loader', options: { minimize: true } }]
    });

    // 3. Don't exclude quill from babel-loader rule
    rules.filter(rule => rule.exclude && rule.exclude.toString() === "/(node_modules|bower_components)/")
        .forEach(rule => rule.exclude = /node_modules\/(?!(quill)\/).*/);
});

// …

mix.js(/* ... */)
   .quill()
```

PR #1471, opened 2018-02-13
hot-module-replacement.md bundled js name is changed

tahaemin:
on this doc, "bundle.js" is wrong. because laravel-mix is bundling js as "app.js". so it was a bit confusing.


tahaemin:
seems no update.

PR #1496, opened 2018-02-25
Update travis environment configuration

vinkla:
We don't need to test against both Linux and macOS. This update will speed up the test suite on travis.

This will reduce the total time with 10 minutes:

screen shot 2018-02-25 at 9 22 37 pm
&nbsp;

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,7 @@
 language: node_js
 
-# Test linux and osx
-os:
-  - linux
-  - osx
-
-# Test against all supported versions of Node.js
 node_js:
-  - "6"
-  - "7"
-  - "8"
-  - "9"
-
-cache:
-  directories:
-    - "node_modules"
+  - 9
+  - 8
+  - 7
+  - 6


thecrypticace:
I think the added value of proving things work on macOS is well worth the time tradeoff. There are definite environmental differences that could cause things to work on macOS and not linux (or vice versa) though _it probably is less likely_ in the case of node projects that don't issue shell commands.

vinkla:
When building tools with Node and Npm we could should expect it to work in all environments supporting Node. I know there have been hickups with Windows in the past but macOS and Linux are both UNIX environments. In the best of worlds Node should work the same way in all environments.

@tpraxl you worked on the initial tests suite setup. Where there a lot of issues with different environments?

tpraxl:
> @tpraxl you worked on the initial tests suite setup. Where there a lot of issues with different environments?

@vinkla I cut most mac builds short, because at this time, mac builds had a major delay. While I set this up, there were no library changes, so this doesn't really tell a story about mac builds being important or not. There are differences between linux and mac builds as far as I can tell from my experience. I cannot tell if these differences are or will be important for laravel-mix. The same is to say about the different node versions. Most of the time, there are no failures for specific versions, but when an error occurs, you appreciate early detection.

To be honest, I'd rather stop supporting the older node versions officially and then remove them from the test setups. Say, we would stop supporting node 5+6, that would make 6 less test runs.

vinkla:
> …but when an error occurs, you appreciate early detection.

I see your point and I agree.

> To be honest, I'd rather stop supporting the older node versions officially and then remove them from the test setups. Say, we would stop supporting node 5+6, that would make 6 less test runs.

I agree, since this is a development tool we could always require the latest version of Node.

Lets leave this up to @JeffreyWay. If he wants to drop tests for older version I can update the PR.

JeffreyWay:
Let's just keep things as they are. I often run into situations where tests will fail on Linux, but not my Mac. Happy to remove older node versions as the usage becomes negligible.

Issue #1524, opened 2018-03-14
Require modules not working at v2.1


dionysiosarvanitis:
- Laravel Mix Version: 2.1.0
- Node Version:  9.2.0
- NPM Version:  5.6.0
- OS: Windows 10

### Description:

After upgrading to 2.1 it seems like Vue registered components are completely ignored outputing errors like "Property or method "xxx" is not defined on the instance but referenced during render".  Registration is done by requiring `.vue` files like below from within app.js

`Vue.component('login', require('./pages/Login.vue'))`

Login.vue
```js
<script>
export default {
    data() {
        return {
            model: {
                identity: '',
                password: '',
                remember: false,
            },
        };
    }
};
</script>
```

When registered inline it works 

```js
Vue.component('login', {
    data() {
        return {
            model: {
                identity: '',
                password: '',
                remember: false,
            },
        };
    }
})
```

Reverting to version 2.0 it works again.

ankurk91:
Here is the workaround
```js
Vue.component('login', require('./pages/Login.vue').default)
```
Something is wrong in vue-loader configs.

ruchern:
`v.2.0`
https://github.com/JeffreyWay/laravel-mix/blob/127ca94cdbe99f9ba21a05d4b2c0d53fc1bddc6a/package.json#L61

`v2.1`
https://github.com/JeffreyWay/laravel-mix/blob/dabc3454d8f5289b7b9a0330a22f2a976cd924b1/package.json#L61

There was a major update to `vue-loader`

ankurk91:
@ruchern 
Any way to restore old behaviour ?
How to return `default` by default?

ruchern:
https://github.com/vuejs/vue-loader/releases

I am looking through this releases to see what are the breaking changes.

ankurk91:
I think we can not restore it
https://github.com/vuejs/vue-loader/blob/master/docs/en/options.md#esmodule

ruchern:
I think OP got to try to try something like

```
import Foo from '@/components/Foo`

Vue.component('foo', Foo)

ankurk91:
Laravel need to remove this old config then

https://github.com/JeffreyWay/laravel-mix/blob/master/src/config.js#L107

ruchern:
And force everyone to go to ES6 modules.

JeffreyWay:
Ugh, I can’t force a breaking change like that for this release. We’ll have to wait for a major release. 

[...]

PR #1526, opened 2018-03-14
Rollback vue-loader to v13

ankurk91:
Laravel mix v2.0.0 was using vue-loader v13.x while v2.1 is using v14.x.
Since vue-loader v14 has a breaking change so lets rollback the version to v13.

https://github.com/vuejs/vue-loader/releases/tag/v14.0.0

Steps taken:
```
npm install
npm uninstall vue-loader
npm install vue-loader@^13
```

Fixes https://github.com/JeffreyWay/laravel-mix/issues/1524

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5458,13 +5458,6 @@
             }
           }
         },
-        "string_decoder": {
-          "version": "1.0.1",
-          "bundled": true,
-          "requires": {
-            "safe-buffer": "5.0.1"
-          }
-        },
         "string-width": {
           "version": "1.0.2",
           "bundled": true,
@@ -5472,6 +5465,13 @@
             "code-point-at": "1.1.0",
             "is-fullwidth-code-point": "1.0.0",
             "strip-ansi": "3.0.1"
+          }
+        },
+        "string_decoder": {
+          "version": "1.0.1",
+          "bundled": true,
+          "requires": {
+            "safe-buffer": "5.0.1"
           }
         },
         "stringstream": {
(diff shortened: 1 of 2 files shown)


ruchern:
Instead of going backwards, should we go forward instead? I'll try to push a PR to fix `vue-loader` on v14 during the weekends!

ankurk91:
I think one can still install vue-loader v14 and laravel mix will start using it.
If we go for v14 then we need a major version bump (v3) in Laravel mix.

ruchern:
I reckon `v2.2` is fine since it's probably just a "feature" patch. I'm basing the version off under SemVer.

JeffreyWay:
I think it probably has to be part of v3. v2.2 would break Laravel apps out of the box.

We'll do it as part of that webpack v4 updates maybe.

ruchern:
Now that sounds like a good idea.

rodrigopedra:
I spent some time last night fixing my app for this change, should have the idea to roll back vue-loader as I managed to find it was the problem.

I was getting the "render function or template not found" error, first I changed global component registration as bellow:

~~~js
// Vue.component('MyComponent', require('./MyComponent.vue')); // BEFORE
Vue.component('MyComponent', require('./MyComponent.vue').default); // AFTER, note the .default
~~~

And local component registration to:

~~~vue
<script>
// let ChildComponent = require('./ChildComponent.vue'); // BEFORE
let ChildComponent = require('./ChildComponent.vue').default; // AFTER, note the .default

export default {
    components: { ChildComponent },
    // ...
}
</script>
<template>
<!-- nothing changed here -->
</template>
~~~

It worked! But I felt it was a little messy to add the default to every component registration, I has about 50 of them so I changed every file to use ES6 `import`s as it was the preferred way per vue-loader docs.

With this rollback this morning I feared having to rollback all this refactoring, but by using ES6 `import` it worked on both vue-loader versions.

Maybe it is worth noting this fixes in a future migration notes on the Laravel Mix 3 release.

JeffreyWay:
@rodrigopedra You don't have to rollback all your refactoring, if you don't want to. You can opt into the new way by adding `Config.vue.esModule = true` to your `webpack.mix.js` file.

```js
const mix = require('laravel-mix');

Config.vue.esModule = true;

mix.js('resources/assets/js/app.js', 'js');
```

ruchern:
I have submitted a PR in #1536 @JeffreyWay @ankurk91 can you guys kindly have a look.

PR #1535, opened 2018-03-16
Fix adding of vue webpack rules when using TypeScript component

jbbr:
When using only `.ts()` component but never `.js()` the vue rules where not added as the TypeScript component overwrites the `webpackConfig` method of the JavaScript component and never calls the parent/super method.

This PR adds a super call to fix this issue.

### DIFF
--- a/src/components/TypeScript.js
+++ b/src/components/TypeScript.js
@@ -35,6 +35,8 @@ class TypeScript extends JavaScript {
      * @param {Object} config
      */
     webpackConfig(config) {
+        super.webpackConfig(config)
+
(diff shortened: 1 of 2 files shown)


jbbr:
hot fix/workaround until this is merged:

```js
const TypeScript = require('laravel-mix/src/components/TypeScript');

mix.extend('ts', new class extends TypeScript {
  webpackConfig(config) {
    this.vue.webpackConfig(config);
    super.webpackConfig(config)
  }
})
```

thanks for implementing the extension interface!

JeffreyWay:
Thank you!

PR #1542, opened 2018-03-19
fix WebpackConfig.js publicPath

### LINKED ISSUE #863: Chunk loading broken due to wrong publicPath
- Laravel Mix Version: 0.12.1
- Node Version 8.0.0
- NPM Version 5.0.x
- OS: Mac OSX 10.9.6 / Ubuntu 16.04.1

### Description:
In https://github.com/JeffreyWay/laravel-mix/pull/209 the webpack public path got changed from './' to just an empty string. As I just recently started to use chunk splitting, i never noticed this before until I wanted to update the main server today as I am using HMR on my local machine.

The issue seems to be that I'm either neither able to set the public path directly, nor the global.entry.base.

The fix is to just set the publicPath '/' as default again (https://github.com/JeffreyWay/laravel-mix/blob/v0.12.1/src/Mix.js#L69):

```js
publicPath: options.hmr ? (http + '://localhost:8080/') : '/'
```

The URLs requested in the browser look like the following:
- **Before:** `my-site.com/som/sub/js/6.(hash).js`
- **After:** `my-site.com/js/6.(hash).js`

~I'm not sure if versioning the files has to do anything with this.~
**Edit:** It doesn't.

### Steps To Reproduce:
1. Regular laravel installation.
2. Trigger the chunk-loading, e.g. `Vue.component('test', r => require(['./test.vue'], r))`
3. Build using `npm run dev` or `npm run production`
4. Go to a different route than home and see the error in the browser

raniesantos:
closes #863 

See [this comment](https://github.com/JeffreyWay/laravel-mix/issues/863#issuecomment-366662394) for context.

### DIFF
--- a/src/builder/WebpackConfig.js
+++ b/src/builder/WebpackConfig.js
@@ -70,7 +70,7 @@ class WebpackConfig {
                   ':' +
                   Config.hmrOptions.port +
                   '/'
-                : ''
+                : '/'
         };
 
         return this;
--- a/test/features/javascript.js
+++ b/test/features/javascript.js
@@ -51,7 +51,7 @@ test.serial('basic JS compilation config.', t => {
             path: path.resolve('test/fixtures/fake-app/public'),
             filename: '[name].js',
             chunkFilename: '[name].js',
-            publicPath: ''
+            publicPath: '/'
         },
         webpackConfig.output
     );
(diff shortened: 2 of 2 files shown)


raniesantos:
And now we wait.

PR #1544, opened 2018-03-19
Output webpack chunks in public/js/

raniesantos:
The current default behavior of laravel-mix when handling code splitting is that it drops the output chunks directly into the public folder. (`public/0.js`)

This modifies the config to place chunks into the js folder with this filename format: (`public/js/0.chunk.js`).


JeffreyWay:
What if there is either no "public/js" directory, or the user uses "public/javascripts" instead?

raniesantos:
Ok fine, I was just assuming everybody used the default folders anyway.

raniesantos:
I just remembered this can actually be used outside of Laravel, it's just so closely tied to the framework.

PR #1552, opened 2018-03-20
We should not use prototypes unless defineProperty is used

lukepolo:
This works because `defineProperty` is not `enumerable` by default


JeffreyWay:
Thanks! This helps a bunch.

PR #1557, opened 2018-03-22
Let the lazy-load chunk have hash for versioning

ItaloBC:
Otherwise, chunks for lazy-loading never get hashed, so it defeats the purpose of have versioning.


JeffreyWay:
This would lead to both querystring and file-based versioning at the same time.

PR #1568, opened 2018-03-28
Make webpack built notifications transient on linux

hailwood:
Currently if you are rebuilding often your notification panel gets flooded with notifications forcing you to either "clear all" or clear the 200 odd "Build Successful" notifications.

This patch marks the notifications as transient on linux meaning they are never saved to the notification center and simply disappear into the abyss after being displayed.


carestad:
This is what I have been looking for! Awesome! When will it be released?

carestad:
@JeffreyWay Any chance of a new release soon?

PR #1570, opened 2018-03-28
Disable CSS & JS minification by default and allow minify() in development

gillisandrew:
This pull request disables the automatic minification of CSS files by css-loader (css-nano) in development and production and removes automatic JS minification in production. Minification should only occur when calling `mix.minify()` or `mix.combine()` are called and matches the standard naming convention for minified versions of assets (`*.min.ext`)

The current setup is far too opinionated about minifying assets in production and offered no way to configure or disable cssnano. This is a real headache when debugging issues in minification such as https://stackoverflow.com/questions/49516072/wrong-output-on-npm-run-prod

### DIFF
--- a/docs/concatenation-and-minification.md
+++ b/docs/concatenation-and-minification.md
@@ -39,7 +39,12 @@ mix.minify(['this/one.js', 'and/this/one.js']);
 There are a few things worth noting here:
 
 1. This method will create a companion `*.min.ext` file. So minifying `app.js` will generate `app.min.js`.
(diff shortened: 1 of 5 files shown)


JeffreyWay:
Thanks,  but I'm not going to disable automatic minification for production JS builds when the huge majority of developers will want precisely that.

gillisandrew:
Minifying is still possible... This is just an effort to make mix more deterministic. The differences in running in development or production environment should be encapsulated in webpack.mix.js, not hidden in the tool. Generating different assets in different environments without being configured to do so is asking for trouble.

PR #1578, opened 2018-04-03
Recognizes scoped package dependencies in mix extension - resolves #1577

### LINKED ISSUE #1577: Bug: Scoped package dependency in mix extension is not recognized.
- Laravel Mix Version: laravel-mix@2.1.11
- Node Version: v8.10.0
- NPM Version: 5.7.1
- OS: Fedora 27

### Description:
I'm trying to build a custom extension for laravel-mix which has few dependencies. These dependencies are scoped packages (i.e they begin with '@' symbol) e.g. `@fortawesome/fontawesome`.

Mix fails to recognize these scoped packages were already installed during the first run of the custom extension. So it tries to re-install them on consecutive run and gets stuck at this stage as I have set `this.requiresReload` to `true`.

> I have found the cause for this bug and will send a PR soon.

### Steps To Reproduce:
1. Create sample mix extension with scoped packages
```
let mix = require('laravel-mix');

class FontAwesome {
    dependencies () {
        this.requiresReload = true;

        return [ '@fortawesome/fontawesome' ];
    }
};

mix.extend('fontAwesome', new FontAwesome());
```
2. Add custom extension to the mix build chain in webpack.mix.js
```
require('<custom-extension-name>');

mix.js('resources/assets/js/app.js', 'public/js')
    .sass('resources/assets/sass/app.scss', 'public/css')
    .fontAwesome();
```
3. Run Mix using `npm run dev`

> It logs `Additional dependencies must be installed. This will only take a moment.` and installs the necessary dependencies (in this case `@fortawesome/fontawesome`)
> Then asks to re-run Mix. This is expected behaviour thus far.

4. But after re-run `npm run dev`

> It still cannot find the scoped package and tries to install it again which is unnecessary.
> It again asks to re-run Mix as I have set `this.requiresReload = true`. This creates a loop and never gets past this stage of loading dependencies.

SaurabhKharivale:
Previous regex pattern replaces all characters after '@' symbol.
So package names like `@fortawesome/fontawesome` would be converted to an empty string.
Resulting in failure to detect the installed dependencies.

### DIFF
--- a/src/Verify.js
+++ b/src/Verify.js
@@ -77,7 +77,7 @@ class Verify {
         list
             .reject(dependency => {
                 try {
-                    return require.resolve(dependency.replace(/@.+$/, ''));
+                    return require.resolve(dependency.replace(/(?!^@)@.+$/, ''));
                 } catch (e) {}
             })
             .tap(dependencies => {
@@ -96,7 +96,7 @@ class Verify {
         if (argv['$0'].includes('ava')) return;
 
         try {
-            require.resolve(name.replace(/@.+$/, ''));
+            require.resolve(name.replace(/(?!^@)@.+$/, ''));
         } catch (e) {
             installDependencies(name, abortOnComplete);
         }


SaurabhKharivale:
I have created https://regexr.com/3na0g with few sample package names for reviewing purpose.

JeffreyWay:
Great, thanks!

PR #1631, opened 2018-05-08
Added babel-plugin-syntax-dynamic-import

lkmadushan:
This plugin helps to load Vue components asynchronously when needed.

See: https://router.vuejs.org/en/advanced/lazy-loading.html


stephan-v:
This should not be included by default because most likely the majority of users will not use this. It is trivially easy to add this your own project, therefore I see no upside to include this optional dependency.

Issue #1433, opened 2018-01-24
npm run prod with purifyCss: true results in completely broken styles


secondfry:
- Laravel Mix Version: 2.0.0 (same problem with ^1.0.0)
- Node Version: v9.4.0
- NPM Version: 5.6.0
- OS: Windows 10

### Description:
`purifyCss` doesn't like minified input. Setting it to true and running `npm run prod` results in styles being lost.

1. `prod` with `purifyCss: false` – https://pastebin.com/B8CEseJM
2. `prod` with `purifyCss: true` or `purify-css` in console – https://pastebin.com/fTCGc7PV
3. `dev` with `purifyCss: true` – https://pastebin.com/5WAFmJcA
4. `dev` with `purifyCss: false` and `purify-css` with `minify: true` afterwards – https://pastebin.com/8HrHyCDE

Search for `.page-title-small`. It's present in 1 (as there was no purify), 3 and 4 (as it should!), but not in 2.

### Steps To Reproduce:
```
mix.options({purifyCss: true})
$ npm run prod
```

### Proposed solution:
Don't minify CSS if `purifyCss` set to `true`, instead pass `minify: true` to `purifyCss` itself.

secondfry:
My temporal fix for this is edit `src/builder/webpack-plugins.js`:
```js
    // Add some general Webpack loader options.
    plugins.push(new webpack.LoaderOptionsPlugin({
        minimize: false, // FIXME Mix.inProduction(),
        options: {
            context: __dirname,
            output: { path: './' }
        }
    }));
```
Even with `minimize: false`, `npm run prod` results in minified files due to `uglifyjs` and `purify-css`.

secondfry:
Also reporting this upstream to `purify-css`.

pxslip:
I just ran into this issue this morning, not sure what others are doing to deal with this, but I have opened #1634 with a potential permanent fix.

secondfry:
That should fix it, I guess.

PR #1634, opened 2018-05-09
Fix issues with minified css and purify

pxslip:
Leave minification to the Purify Webpack plugin if it is in use
Add a test to ensure the minify value is being set properly

This relates to the issue brought up in #1433


pxslip:
Feel free to let me know if there are any changes or additional tests that would help.

PR #1665, opened 2018-06-02
Support port number from CLI

morrislaptop:
If I run `npm run hot -- --port=1235`, the the `public/hot` file is still `http://localhost:8080`

This PR fixes this so the url in `public/hot` is `http://localhost:1235`


morrislaptop:
* bump

PR #1667, opened 2018-06-04
Don't try to minify code in CopyFilesTask tasks

### LINKED ISSUE #1666: `mix.copy();` tries to minify JavaScript (includes reduced test case)
- Laravel Mix Version: 2.1.11 (`npm list --depth=0`)
- Node Version (`node -v`): 8.1.3
- NPM Version (`npm -v`): 5.0.3
- OS: macOS 10.13.4

### Description:

When using `mix.copy();` to copy a JavaScript file, Mix tries to minify the file if NODE_ENV=production. If it hits ES2015 syntax, Mix throws an error and recommends using `mix.babel()`.

### Steps To Reproduce:

1. Follow instructions for standalone install of Mix verbatim, including making an empty directory and copying the mix file and npm scripts.

2. Modify the mix file such that the only mix directive is:

`mix.copy('src/app.js', 'dist/app.js');`

3. Create a file, `src/app.js`, and put the following line of code in it:

`var test = () => console.log('hello');`

4. `npm run production`

blakewatson:
Fixes #1666. Two things:

1. I'm pretty sure `mix.copy()` was intended to simply move files while leaving their contents untouched. But correct me if I'm wrong.
2. I don't know how to write a test for this.

### DIFF
--- a/src/webpackPlugins/CustomTasksPlugin.js
+++ b/src/webpackPlugins/CustomTasksPlugin.js
@@ -47,9 +47,9 @@ class CustomTasksPlugin {
      * Minify the given asset file.
      */
     minifyAssets() {
-        let tasks = Mix.tasks.filter(
-            task => task.constructor.name !== 'VersionFilesTask'
-        );
+        let tasks = Mix.tasks.filter(task => {
+            return task.constructor.name !== 'VersionFilesTask' && task.constructor.name !== 'CopyFilesTask';
+        });
 
         tasks.forEach(task => {
             task.assets.forEach(asset => {


src/webpackPlugins/CustomTasksPlugin.js
```diff
@@ -47,9 +47,9 @@ class CustomTasksPlugin {
      * Minify the given asset file.
      */
     minifyAssets() {
-        let tasks = Mix.tasks.filter(
-            task => task.constructor.name !== 'VersionFilesTask'
-        );
+        let tasks = Mix.tasks.filter(task => {
+            return task.constructor.name !== 'VersionFilesTask' && task.constructor.name !== 'CopyFilesTask';
```
dbpolito: i simpler way for this this check:

```js
let tasks = Mix.tasks.filter(
    task => ! ['CopyFilesTask', 'VersionFilesTask'].includes(task.constructor.name)
);
```

or just add the new check without changing the ecma:

```js
let tasks = Mix.tasks.filter(
    task => task.constructor.name !== 'VersionFilesTask' && task.constructor.name !== 'CopyFilesTask'
});
```

dbpolito:
This is a huge improvement for me, i would really appreciate this getting merged.

My build time goes from `150s` to `50s` with this change, as i have some pretty big compiled/minified files.

src/webpackPlugins/CustomTasksPlugin.js
blakewatson: @dbpolito The top one is nice. Love that.

PR #1669, opened 2018-06-05
Allow autoloading of module properties

josh-taylor:
This PR allows the autoloading of module properties as shown in the [ProvidePlugin docs](https://webpack.js.org/plugins/provide-plugin/#usage-lodash-map)

It adds support the following syntax in the autoload method:

```js
mix.autoload({
    'lodash.map': '_map'
});
```

Which would then allow you in your source code to do this:

```js
_map([1, 2, 3), () => {});
```

It would be the equivalent of doing:

```js
let { map } = require('lodash');

map([1, 2, 3], () => {});
```

### DIFF
--- a/src/components/Autoload.js
+++ b/src/components/Autoload.js
@@ -10,7 +10,7 @@ class Autoload {
 
         Object.keys(libs).forEach(library => {
             [].concat(libs[library]).forEach(alias => {
-                aliases[alias] = library;
+                aliases[alias] = library.includes('.') ? library.split('.') : library;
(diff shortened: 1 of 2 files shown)


ankurk91:
@josh-taylor 

This PR has broken this code
```js
mix.autoload({
    'jquery/dist/jquery.slim': ['$', 'window.jQuery', 'jQuery'],
  })
```
It was working before this change. Now browser is throwing errors like - `$ is not defined`.

PR #1671, opened 2018-06-06
Add inDevelopment method

driesvints:
This method comes in handy when you want to check if you're working in a development environment. As production and development are the most common environments to work in, often more names like staging, testing, etc. are used to denounce specific environments. Having a specific helper to indicate the development environment can come in handy when running certain processes which are only necessary for development.

### DIFF
--- a/src/Api.js
+++ b/src/Api.js
@@ -95,6 +95,13 @@ class Api {
     inProduction() {
         return Mix.inProduction();
     }
+
+    /**
+     * Helper for determining a development environment.
+     */
+    inDevelopment() {
+        return Mix.inDevelopment();
+    }
 }
 
 module.exports = Api;
--- a/src/Mix.js
+++ b/src/Mix.js
@@ -34,6 +34,13 @@ class Mix {
     }
 
     /**
+     * Determine if Mix is executing in a development environment.
+     */
+    inDevelopment() {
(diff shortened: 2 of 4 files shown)


stephan-v:
Why would you not simply use:

`!Mix.inProduction()`

This seems a bit unneccesary especially since I doubt this will get used a lot.

driesvints:
@stephan-v the opposite of `production` is not `development`. `development` is a specific environment for your Laravel Mix configuration and it's explicitly defined in your `package.json` for local development. You can add more environments like `staging`, `ci`, `testing`, etc. 

For example: you could have a `staging` environment which acts like `production` but has a few nuances. However, if you would use `!Mix.inProduction()` to target your local `development` environment, your `staging` environment would be targeted as well, which you wouldn't want.

Since `development` and `production` are two default environments which ship with Laravel Mix, it makes sense to have methods which can target them. That way you can use those to manipulate the behavior of your scripts.

JeffreyWay:
Sorry for the long delay...

But if I ran mix.inDevelopment() in my local environment, it would by default return false...which is confusing.

Let's wait and see if we have more requests for this. After a few years, yours is the only PR to change this, which makes me think it's not overly necessary.

driesvints:
No problem at all! :)

PR #1726, opened 2018-08-01
Parse .babelrc as JSON5, as does babel-core

hackel:
.babelrc files support the JSON5 specification.

See: http://babeljs.io/docs/en/babelrc/


ankurk91:
My favorite  feature is comments in .json file

tianyong90:
Perhaps it's better to upgrade babel, babel7 supports using `babel.config.js` . 😄

JeffreyWay:
Agree with @tianyong90. Let's keep things as they are until we bump to babel7 soon.

PR #1743, opened 2018-08-17
Embedded styles for HMR mode, resolved css injecting problem

FlexLi99:
All styles are extracting now, but when I'm starting hmr mode and changing my styles, I have to refresh my browser each time. Is there any other way to solve the problem?


JeffreyWay:
After four months of testing this PR daily, I've decided to merge it. 


:)

FlexLi99:
Ooo! Great news :)) @JeffreyWay thank you so much!!!

PR #1824, opened 2018-11-24
Expose the OS system notification timeout to Laravel Mix Config

connecteev:
Expose the OS system notification timeout to Laravel Mix Config, so that you can control the timeout from the Mix config


JeffreyWay:
I'd rather not make every possible thing configurable. If there's a more sensible setting for the timeout, feel free to submit a PR to override that.

connecteev:
Makes sense. I just feel like 5 sec is a very long timeout, I find myself waiting for the OS notification to go away quite a bit. Perhaps change that to 2 sec?

Issue #1118, opened 2017-08-16
manifest.js is not created in the same folder as vendor.js


yhosun:
- Laravel Mix Version: 1.4.2
- Node Version (`node -v`): v6.10.3
- NPM Version (`npm -v`): 3.10.10
- OS: OSX

### Description:

After upgrading to Laravel Mix 1.4.2 from 0.11.4, `manifest.js` is not created in the same folder as `vendor.js`.  Instead, `manifest.js` was added to the folder where a javascript is built by `js()` function.

### Steps To Reproduce:
**webpack.mix.js**:
```js
mix.setPublicPath('app/public')
    .js('asset/js/test/test.js', 'js/test')
    .extract(['vue'], 'js/vendor.js');
```
**webpack output**:
```
       /js/vendor.js   82.4 kB       0  [emitted]  /js/vendor
    /js/test/test.js  72 bytes       1  [emitted]  /js/test/test
/js/test/manifest.js   1.38 kB       2  [emitted]  /js/test/manifest
```
I believe it was `/js/manifest.js` instead of `/js/test/manifest.js` in  Laravel Mix 0.11.4.

etiennellipse:
I upgraded my Laravel Mix version to 1.5.1 today, and this issue reappared! I am the reporter of #553.

If that helps anyone, I reused my old workaround. The idea is to push a CommonsChunkPlugin config that will likely override the one setup by Mix. It seems like it has no effect on the vendor chunk

```
const myScriptBasePath = 'dist/core'; // this is where my manifest will output

mix.js('some/path/myEntry', 'dist/myOutput')
    .webpackConfig({
        plugins: [
            new webpack.optimize.CommonsChunkPlugin({
               name: path.join(myScriptBasePath, 'manifest'),
               minChunks: Infinity
            })
        ]
    })
```

This will output the manifest into the path defined by myScriptBasePath. It seems to work alright, while waiting for the issue to be fixed.

hafezdivandari:
I found a workaround for this issue by creating an empty js file in `js/empty.js` and doing this:

```
mix.setPublicPath('app/public')
    .js('asset/js/test/test.js', 'js/test')
    .js('asset/js/empty.js', 'js')
    .extract(['vue']);
```

**webpack output:**
```
/js/test/test.js  72 bytes       1  [emitted]  /js/test/test
    /js/empty.js   5 bytes       1  [emitted]  /js/empty
[...]

Issue #1695, opened 2018-06-28
Cannot define output path for manifest.js when extracting vendor libraries


iraklisg:
### Description:
In my application I have two entry points (`js/main.js`, `/js/appB/main.js`) that share a common `some-vendor-library` and I want to extract it in a separate file.

The  `webpack.mix.js` file:

```js
mix.js('resources/assets/js/main.js', 'public/js/main.js');
mix.js('resources/assets/js/appB/main.js', 'public/js/appB/main.js');

mix.extract(['some-vendor-library'], 'public/js/vendor.js')
```

Running the code above, will result in successfully extracting the common vendor library into `vendor.js` and create the required `manifest.js` file.

However, although I can explicitly define the output path for `vendor.js` (i.e. `'public/js/vendor.js'`) , I cannot do the same with `manifest.js`. Specifically, `manifest.js` is located into the path defined by the last `mix.js()` call, that is  `public/js/appB/manifest.js` in this particular case.

Is it possible to explicitly define a path for **both** `vendor.js` and `manifest.js` ?

IMHO it would also make sense the `manifest.js` to be automatically placed in the parent directory of `vendor.js`, as it is defined by the the `mix.extract()`

For instance, in case we call 
```
mix.extract(['some-vendor-library'], 'public/js/some-directory/vendor.js')
```
, the manifest to be placed into the `public/js/some-directory/` directory

elliotleelewis:
IMO the second parameter to the extract function should be the directory for both the `vendor.js` and `manifest.js` files. I.E. we should call it as:

```
mix.extract(['some-vendor-library'], 'public/js/some-directory');
```

I wouldn't think there would be many use cases where it would make sense to have `manifest.js` and `vendor.js` in different directories 😄

dvlpr91:
@iraklisg Did you solve it? I have the same problem.

@elliotleelewis I can not solve it your way.

nmfzone:
Not stale. Bumb up..

ccgintranet:
We are also having this same issue. In our case, the last call to `mix.ts` (for TypeScript) will be the path that `vendor.js` & `manifest.js` will be compiled to.

"laravel-mix": "^5.0.1",

Issue #1203, opened 2017-09-19
Using extract(); the manifest.js output path is different from vendor.js (new)


getriebesand:
- Laravel Mix Version: #.#.# (`npm list --depth=0`) 1.4.3
- Node Version (`node -v`): 6.10.0
- NPM Version (`npm -v`): 5.3.0
- OS: Windows7

### Description:
See Bugreport #553
The path ob vendor.js and manifest.js differs.
The path of manifest.js is the same path as path of last processed js file.

### Steps To Reproduce:
mix.setPublicPath('public')
mix
    .extract([
        'vue',
        'lodash',
        'axios'
    ],'js/vendor');
mix.js('resources/assets/js/app.js', 'public/js');
mix.js('resources/assets/js/acp/app.js', 'public/js/acp');

vendor.js is in public/js
manifest.js is in public/js/acp

getriebesand:
duplicate of #1118

goper-leo:
Have you solve this? Im using laravel-mix 1.7.2

PR #1855, opened 2018-12-12
WIP: Add failing test for issue #1118

rtheunissen:
See #1118, which was closed but not resolved.

### DIFF
--- a/test/features/extraction.js
+++ b/test/features/extraction.js
@@ -15,6 +15,15 @@ test.serial('JS compilation with vendor extraction config', t => {
     );
 });
 
+test.cb.serial('vendor extraction manifest in same path as output', t => {
+    mix.js('resources/assets/js/app.js', 'js').extract(['vue'], 'libraries.js');
+
+    compile(t, () => {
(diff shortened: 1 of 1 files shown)


rtheunissen:
I think this should somehow consider the `output` path, if provided: https://github.com/JeffreyWay/laravel-mix/blob/master/src/components/Extract.js#L74

rtheunissen:
This gets tricky with multiple extractions, because they may not share the same base path.

rtheunissen:
Overriding `optimization.runtimeChunk.name` has no effect, so not sure where that path is generated in the first place. Seems like there are multiple unresolved issues about this:
- https://github.com/JeffreyWay/laravel-mix/issues/1118
- https://github.com/JeffreyWay/laravel-mix/issues/1203
- https://github.com/JeffreyWay/laravel-mix/issues/1695

rtheunissen:
We ended up adjusting our public path to accommodate this.

culturegr:
@rtheunissen could you please elaborate on this 
> We ended up adjusting our public path to accommodate this?

It seems that the problem still exists (see #1118), meaning that `manifest.js` is not created in the same directory as `vendor.js`

PR #1901, opened 2018-12-30
Explicitly include only node_modules during auto-extract

### LINKED ISSUE #1900: App modules appear in vendor extraction when passing no args to extract
Using the latest version of mix, we've noticed that some project modules are ending up in our auto-extracted `vendor.js` file. Is the expected behavior instead that only files from node_modules be auto-extracted? What might cause project files to be included?

rtheunissen:
This is an attempt to fix #1900

### DIFF
--- a/src/components/Extract.js
+++ b/src/components/Extract.js
@@ -92,8 +92,11 @@ class Extract {
         // If the user didn't specify any libraries to extract,
         // they likely want to extract all vendor libraries.
         if (Object.keys(config.cacheGroups).length === 0) {
-            config.chunks = 'all';
-            config.name = this.extractions[0].output;
+            config.cacheGroups['vendor'] = {
+                test: new RegExp('[\\/]node_modules[\\/]', 'i'),
+                name: this.extractions[0].output,
+                chunks: 'all'
+            };
         }
 
         return config;


rtheunissen:
We are using this fork currently and are happy to see only node_modules sources in vendor.js - everything else is accounted for as expected. :+1:

src/components/Extract.js
```diff
@@ -92,8 +92,11 @@ class Extract {
         // If the user didn't specify any libraries to extract,
         // they likely want to extract all vendor libraries.
         if (Object.keys(config.cacheGroups).length === 0) {
-            config.chunks = 'all';
-            config.name = this.extractions[0].output;
+            config.cacheGroups['vendor'] = {
+                test: new RegExp('[\\/]node_modules[\\/]', 'i'),
+                name: this.extractions[0].output,
+                chunks: 'all'
+            };
```
rtheunissen: Create a single cache group with an explicit test for `node_modules` in the path.

rtheunissen: I think the case-insensitive regex is overkill here but it's consistent with the logic in `createCacheGroup`. @JeffreyWay let me know if you'd like me to refactor this a bit further.

thecrypticace:
@rtheunissen Could you test the latest Laravel Mix beta and see if this is still an issue? I'm fairly certain it shouldn't be anymore. If it is I'd like to figure out why and fix it.

Issue #1881, opened 2018-12-19
Error when using webp images as background-image css


chriswright-interconnect:
-   Laravel Mix Version: 4.0.7 
-   Node Version v8.12.0:
-   NPM Version 6.4.1:
-   OS: Windows 10

### Description:

In one of my .scss files I use a .webp image in a background property. When running watch/prod/dev I get an error "Module parse failed: Unexpected character '�' (1:4)" and laravel-mix fails. 

### Steps To Reproduce:
- Add a .webp image as the background-image property of an element in your scss. 
- Run 'npm run watch' to get the error.

I've tried using plugins; [image-webpack-loader](https://www.npmjs.com/package/image-webpack-loader) and [webp-loader](https://www.npmjs.com/package/webp-loader) and tried inserting into the webpack.config.js via webpack.mix.js but still get the same error. 

Is there something I'm missing to get this to work? Or can anyone reproduce?

ankurk91:
There is no rule set for `webp`

https://github.com/JeffreyWay/laravel-mix/blob/9acd47220fd671f2b317ccc4b109ad06b5ea78da/src/builder/webpack-rules.js#L13

dementor5:
I have this problem too. :(

chriswright-interconnect:
Sorry I've not replied earlier, been away for the holidays.

@ankurk91 , is there any way to add the rule set to webpack-rules.js?
Something similar to adding this to webpack.mix.js?

```
.webpackConfig({
    module: {
      rules: [{
        test: /(\.(png|jpe?g|gif|webp)$|^((?!font).)*\.svg$)/,
      }]
    }
  })
```

JeffreyWay:
Fixed in #1917

JeffreyWay:
Either pull in the master branch, or wait until we tag a new patch release.

PR #1917, opened 2019-01-07
Handle webp image type

ankurk91:
May be fixes

https://github.com/JeffreyWay/laravel-mix/issues/1881

Read more about webp image format
https://developers.google.com/speed/webp/


JeffreyWay:
Thanks, @ankurk91!

PR #1918, opened 2019-01-07
Switch to babel-preset-react-app

andreyberezhnoy:
What do you think about babel-preset-react-app from https://github.com/facebook/create-react-app?

It supports some useful features like:
- proposal-class-properties
- plugin-transform-react-remove-prop-types


JeffreyWay:
I don't use React, so will need more input from that community before merging this.

andreyberezhnoy:
@JeffreyWay do you have any ideas why tests failed? I just changed the package name, hm..

ruchern:
@andreyberezhnoy tests are broken on Vue cases.

andreyberezhnoy:
@ruchern oh, lol. problem was with npm shrinkwrap. Fixed

ruchern:
What are the differences between `@babel/preset-react-app` and `babel-preset-react-app`?

ruchern:
As I am reading this: https://github.com/facebook/create-react-app/tree/master/packages/babel-preset-react-app I do not see why Laravel-Mix will have to use this since that particular package is right out of the box for `create-react-app`.

andreyberezhnoy:
@ruchern 
babel-preset-react-app supports:
- typescript
- remove prop types from production (for better performance and lower bundle size)
- class properties

azdanov:
This seems like a good default option. [`@babel/preset-react`](https://babeljs.io/docs/en/babel-preset-react) by itself takes care only of jsx.

I'm not familiar with `vue`, but can guess that it does a lot of [similar](https://github.com/JeffreyWay/laravel-mix/pull/1918#issuecomment-457167153) optimizations. With react those need to be added manually.

andreyberezhnoy:
@azdanov also they want to add this awesome plugin to preset https://github.com/jamiebuilds/babel-react-optimize

https://github.com/facebook/create-react-app/pull/6219

azdanov:
@andreyberezhnoy Didn't know about that, looks sweet.

Also, this would work really well with [`mix.react()`](https://laravel-mix.com/docs/4.0/mixjs#react-support) for simplicity sake.

azdanov:
It looks like `babel-preset-react-app` uses `@babel/preset-react` under the hood.

https://github.com/facebook/create-react-app/blob/714599dae652426331007f95cee5e26ea627a1b1/packages/babel-preset-react-app/create.js#L100

So the transition shouldn't cause any issues unless it's some weird edge cases.

@andreyberezhnoy I'm interested in trying out this PR now. Do you know a way to override current settings for react inside `webpack.mix.js`?

andreyberezhnoy:
@azdanov create the `.babelrc` file near `webpack.mix.js`, then add something like than:

```json
{
  "presets": ["babel-preset-react-app"]
}
```

it should works

ecreeth:
@sebastiandedeyne I would like to know your opinion about this

src/components/React.js
```diff
@@ -5,15 +5,15 @@ class React extends JavaScript {
...
+        return ['babel-preset-react-app'];
     }
 
     /**
      * Babel config to be merged with Mix's defaults.
      */
     babelConfig() {
         return {
-            presets: ['@babel/preset-react']
+            presets: ['babel-preset-react-app']
```
ruchern: Based on this [README](https://www.npmjs.com/package/babel-preset-react-app), shouldn't it be 

```presets: ["react-app"]```

sebastiandedeyne:
Not sure if it makes sense for Laravel Mix to support class properties in React apps but not for other frameworks. Keeping Babel config as consistent as possible between projects is kinda what makes Mix interesting.

PR #1925, opened 2019-01-12
Reduce build time by tweaking terser configs

ankurk91:
**Enabling `cache`  for terser (uglify-js) will reduce build time to 50%**

https://github.com/webpack-contrib/terser-webpack-plugin#cache

**Enabling `parallel` option will take advantage of multi-process when available.**

https://github.com/webpack-contrib/terser-webpack-plugin#parallel 

Laravel Mix is already using cache for babel-loader.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -160,6 +160,8 @@ module.exports = function() {
          * @type {Object}
          */
         terser: {
+            cache: true,
+            parallel: true,
             sourceMap: true,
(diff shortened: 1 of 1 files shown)


lkraav:
Yep, 50% increase immediately for `npm run prod`, good job!

JeffreyWay:
Wew! Good find, @ankurk91!

PR #1960, opened 2019-01-25
Allow use dynamic import

andreyberezhnoy:
With this changes we can use dynamic import by default.

https://webpack.js.org/guides/code-splitting/#dynamic-imports

this is useful if you want to split modules into chunks and connect them dynamically only where you need.

for example:
```js
if (document.getElementById('root') {
  import('./module.js').then(module => module.init());
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     ],
     "dependencies": {
         "@babel/core": "^7.2.0",
+        "@babel/plugin-syntax-dynamic-import": "^7.2.0",
         "@babel/plugin-proposal-object-rest-spread": "^7.2.0",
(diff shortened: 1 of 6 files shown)


marwan2:
I was trying many times to use dynamic import, 
and finally recognize that you aren't enabling it !!!!

Did you release it into production ?

msonowal:
@JeffreyWay 
Please check this which will give people to ability to upgrade their projects to v4 from v3

pmochine:
I'm curious why it's not accepted. I know you want to wait for webpack 5, but there is no release date for it and it's april...

PR #1974, opened 2019-02-01
Support import .vue SFC from node_modules

renatofdds:
Currently JavaScript component excludes all imports from node_modules for transformation with babel. This breaks .vue Single File Components (SFC) imported from node_modules with script blocks (which through vue-loader generates new modules imports as '/node_modules/[..].vue.js').


ankurk91:
Essentially package authors should supply complied component.

PR #1995, opened 2019-02-13
Fixed HMR on Windows

edcoreweb:
Currently on Windows HMR does not work:

How to reproduce:
- delete js/css outputs
- run `npm run hot`

You will get 404 errors.

The following section: `path.resolve(Mix.isUsing('hmr') ? '/' : Config.publicPath)` 
will result on `D:\\` instead of the expected `/`, where `D` is the windows drive letter.

Expected: (linux, mac)
```js
output: {
  path: '/',
  filename: '[name].js',
  chunkFilename: '[name].js',
  publicPath: 'http://localhost:8080/'
}
```

Windows:
```js
output: {
  path: 'D:\\',
  filename: '[name].js',
  chunkFilename: '[name].js',
  publicPath: 'http://localhost:8080/'
}
```

This fixes: #1993

PR #2000, opened 2019-02-14
Generate css files with HMR

edcoreweb:
## What is the problem ?
#1743 Introduced hot-reload for styles, but in the process **broke** the generation of the **css files**.

Worked before with `--hot`:
- `<link href="{{ mix('css/app.css') }}"`

The above example gives a **404** now.

## Why is it a problem ?

- you have critical css which needs to be loaded before the js.
- you make js size calculation on dom elements, the hot styles will kick in too late

## How is it fixed ?

The styles are always (with/without `--hot`) passed through **ExtractTextPlugin**

If `--hot` is passed the output of the  **ExtractTextPlugin** will not be removed, but passed to the  **style-loader**, otherwise it will be removed.
```js
const applyLoaders = (hmr, loaders) => {
  loaders = extractPlugin.extract({
    fallback: 'style-loader',
    use: loaders,
    remove: !hmr
  })

  return hmr ? ['style-loader', ...loaders] : loaders
}
```

Follow up of #1995

### DIFF
--- a/src/components/Preprocessor.js
+++ b/src/components/Preprocessor.js
@@ -88,14 +88,19 @@ class Preprocessor {
                     });
                 }
 
+                const applyLoaders = (hmr, loaders) => {
+                    loaders = extractPlugin.extract({
+                        fallback: 'style-loader',
+                        use: loaders,
+                        remove: !hmr
+                    })
+
+                    return hmr ? ['style-loader', ...loaders] : loaders
+                }
+
                 rules.push({
                     test: preprocessor.src.path(),
-                    use: Mix.isUsing('hmr')
-                        ? ['style-loader', ...loaders]
-                        : extractPlugin.extract({
-                              fallback: 'style-loader',
-                              use: loaders
-                          })
+                    use: applyLoaders(Mix.isUsing('hmr'), loaders)
                 });
 
                 this.extractPlugins = (this.extractPlugins || []).concat(


lmeysel:
Thanks for that - was searching sooo long for a solution. Until the PR is merged, i wrote a workaround, to avoid modifying the package-code within node_modules. Maybe this helps someone...

```javascript
mix.extend('_hmrwinfix', new class {
	webpackConfig(config) {
		if (!mix.config.hmr)
			return;

		if (require('laravel-mix/package.json').version !== '4.0.15')
			console.warn('\x1b[43m CAUTION: This fix is written for laravel-mix@4.0.15 - for later versions, check whether the PR #1995 and #2000 are merged - otherwise this fix may cause misbehavior. \x1b[0m');

		// https://github.com/JeffreyWay/laravel-mix/pull/1995
		config.output.path = '/';

		// https://github.com/JeffreyWay/laravel-mix/pull/2000
		let extractPlugin = config.plugins.filter(p => p.filename && /\.css$/.test(p.filename));
		config.module.rules.forEach(r => {
			if (typeof r.test === 'string' && /\.(s[ac]ss|css)$/.test(r.test))
				r.use = ['style-loader', ...extractPlugin.shift().extract({
					fallback: 'style-loader',
					use: r.use.splice(1),
					remove: false
				})];
		});
	}
})._hmrwinfix();
```

edcoreweb:
@lmeysel 

I used this neat package https://github.com/ds300/patch-package which applies a git patch after `npm install`

The git patch is versioned and it looks like this: http://prntscr.com/mzebez

Issue #2008, opened 2019-02-21
laravel-mix make friends with vue <script lang="tsx">


romanmarkelov:
-   Laravel Mix Version: 4.0.14

I get an error if I declared vue component like below
```vue
// World.vue
<script lang="tsx">
import Vue, { CreateElement, VNode } from 'vue'
import { Component } from 'vue-property-decorator'

@Component
export default class World extends Vue {
  render (h: CreateElement): VNode {
    return <p>This is rendered via TSX</p>
  }
}
</script>
```
```ts
// tsconfig.json
{
  "compilerOptions": {
    "target": "esnext",
    "lib": [
      "dom",
      "esnext"
    ],
    "module": "es2015",
    "moduleResolution": "node",
    "experimentalDecorators": true,
    "strict": true,
    "noUnusedLocals": true,
    "noUnusedParameters": true,
    "jsx": "preserve",
    "jsxFactory": "h"
  },
  "include": [
    "./**/*.ts"
  ],
  "types": [
    "vue"
  ],
  "compileOnSave": false
}
```
```json
// .babelrc
{
  "presets": [
    ["@babel/env", { "modules": false }]
  ],
  "plugins": [
    "@babel/syntax-jsx",
    "transform-vue-jsx"
  ]
}
```
```js
// mix
mix.webpackConfig({
    module: {
      rules: [{
          test: /\.tsx?$/,
          exclude: /node_modules/,
          use: [
            'babel-loader',
            {
              loader: 'ts-loader',
              options: {
                appendTsSuffixTo: [/\.vue$/],
                appendTsxSuffixTo: [/\.vue$/],
              }
            }
          ]
        },
      ]
    }
  })
```
```
// error
ERROR in ./src/components/World.vue?vue&type=script&lang=tsx& (../node_modules/babel-loader/lib!../node_modules/ts-loader??ref--0-1!../node_modules/vue-loader/lib??vue-loader-options!./src/components/World.vue?vue&type=script&lang=tsx&)
Module build failed (from ../node_modules/babel-loader/lib/index.js):
SyntaxError: /vue-class-component/example/src/components/World.vue: Unterminated regular expression (15:15)

  13 |         rendered;
  14 |         via;
> 15 |         TSX < /p>;
     |                ^
  16 |     }
  17 | };
```
@ktsn proposed [solution](https://github.com/vuejs/vue-class-component/issues/314)

[...]

PR #2009, opened 2019-02-21
Fix append suffix for ts

romanmarkelov:
More info https://github.com/JeffreyWay/laravel-mix/issues/2008


romanmarkelov:
I think better solution add new component tsx or use mix.extend

PR #2089, opened 2019-05-02
fix: remove loaderOption (migration is done)

cmizzi:
As `svelte` can't work using `laravel-mix`, the usage of `loaderOption` seems deprecated is our case (webpack 4) and doesn't need to pass options to every loader anymore


laracasts:
Thanks!

PR #2102, opened 2019-05-14
Allow manifest files to handle sub directory installation

saulhudson:
If you have Laravel installed in a sub directory from your server root directory (e.g. /var/www/blog/) the mix-manifest.js file doesn't correctly hold the path to your assets because it omits the sub directory. Additionally setPublicPath and setResourceRoot are used in other places so those aren't options for simply prepending a path to the right hand side of the mix-manifest file.

e.g.
> "/js/jquery.min.js": "/js/jquery.min.js?id=220afd743d9e9643852e", 

should read

> "/js/jquery.min.js": "/blog/js/jquery.min.js?id=220afd743d9e9643852e", 

Using the code in the pull request setting the new config option is as simple as
``mix.setRootSubdirectory('blog'); 

I think this would be useful to anyone running Laravel in a subdirectory on their web server or development environments.

### DIFF
--- a/src/Api.js
+++ b/src/Api.js
@@ -46,7 +46,18 @@ class Api {
 
         return this;
     }
+    
+    /**
+     * Set a root directory for a project in a subdirectory
+     *
+     * @param {string} defaultPath
+     */
+    setRootSubdirectory(defaultPath) {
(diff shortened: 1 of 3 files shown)


laracasts:
I'm worried this is too confusing. I'd rather fix the issue elsewhere so that we don't have to add another method like this. I'll take a look.

PR #2111, opened 2019-05-19
Cross-platform yarn detection

kristianheljas:
`command -v yarn >/dev/null` works only on Unix systems.
What about just to execute yarn itself to detect its presence?

### DIFF
--- a/src/Dependencies.js
+++ b/src/Dependencies.js
@@ -71,7 +71,7 @@ class Dependencies {
 
         if (!forceNpm) {
             try {
-                childProcess.execSync('command -v yarn >/dev/null');
+                childProcess.execSync('yarn -v');
 
                 return `yarn add ${dependencies} --dev --production=false`;
             } catch (e) {}


QWp6t:
Add `{ stdio: 'ignore' }` as 2nd parameter to suppress stderr.

```node
childProcess.execSync('yarn -v', { stdio: 'ignore' });
```

This is fine because exit code !== 0 will still throw an error.

Here's a reduced test case to prove it works.

```node
const childProcess = require('child_process');

try {
  childProcess.execSync('yarn -v', { stdio: 'ignore' });
  console.log('yarn detected');
} catch (_unused) {
  console.log('yarn NOT detected');
}
```

JeffreyWay:
@QWp6t Can you confirm that your code example works for both Mac + Windows users? If so, I'll merge it in.

QWp6t:
I can confirm that my example above works in Windows and Ubuntu, but unfortunately I don't have macOS to test in that environment.

kristianheljas:
Hey, I see no reason why this should fail in MacOS since using builtin, which should work the same across platforms.

I found that I can actually detect the package manager used for the command via $npm_execpath.
Test script: `"env":  "echo $npm_execpath"`

Tested in Windows and Ubuntu, got following paths:
```
C:\Program Files\nodejs\node_modules\npm\bin\npm-cli.js
C:\Program Files (x86)\Yarn\bin\yarn.js

/usr/lib/node_modules/npm/bin/npm-cli.js
/usr/lib/node_modules/yarn/bin/yarn.js
```

I could rewrite the detection to search `"yarn"` from `process.env.npm_execpath` with fallback to the previous method.

Something like:
```
try {
    if (process.env.hasOwnProperty('npm_execpath') === false || process.env.npm_execpath.indexOf('yarn') === -1)
    {
        childProcess.execSync('yarn -v',{ stdio: 'ignore' });
    }
    return `yarn add ${dependencies} --dev --production=false`;
} catch (e) {}
```

This way:
1. Using npm/yarn will autodetect without spawning child process
2. Calling directly will prefer yarn if present

@JeffreyWay Does this sound reasonable?

Also, I cannot see how `forceNpm` can be triggered, `Assert.dependencies` will never assign it and it defaults to `false`. For those people who call it directly and prefer npm but have both installed :)

ul8:
Just because someone has installed yarn doesn't mean they're using it. This should be quite common now that some (probably many) people are migrating back to npm: Want to use npm, but still have yarn installed for older projects etc. 

Can't we instead check if there is a yarn.lock in current directory? This would confirm it is actually in use. And exposing the forceNpm option would be a nice extra.

kristianheljas:
@ul8 My previous comment resolves that issue, haven't gotten feedback for it yet.

As this is about child processes, I would automatically detect if main process is npm or yarn so no need to check the lock file.

mikeerickson:
See https://github.com/JeffreyWay/laravel-mix/issues/2196 as this PR would fix my issue as well

mikeerickson:
Actually, does this PR work if both `yarn` is installed, but not the desired package manager.  I am thinking that if `yarn.lock` file does not exist in project, use `npm`

kristianheljas:
@mikeerickson My comment above explains a solution to detect if yarn or npm was run, current PR indeed always prefers Yarn if avail.

But there is no response from @JeffreyWay so I figure we won't get it solved anytime soon.

QWp6t:
RIP `yarn`

https://github.com/JeffreyWay/laravel-mix/commit/adf6524efddf1f8a5ef659e4bf801ba0adbbe369

JeffreyWay:
Yeah, I stripped Yarn entirely. I don't see much reason to maintain support for both at this point.

PR #2113, opened 2019-05-23
Adding a config to check if mix is running in a testing env

lukzgois:
## Why?

I've ran into some problems using the `.extract()` feature while running my frontend test suite. The best way to fix it is by disabling the extraction while running the tests:

```js
"test": "cross-env NODE_ENV=testing mochapack --webpack-config=node_modules/laravel-mix/setup/webpack.config.js --require spec/frontend/setup.js spec/frontend/\\*\\*/\\*.spec.js --growl",
```

```js
if (process.env.NODE_ENV !== 'testing') {
  mix.extract()
}
```

This PR adds a helper function to determine if Mix is running in the test env, so we can rewrite the above function as:

```js
if (! mix.inTesting()) {
  mix.extract()
}
```


laracasts:
Let's stick with process.env.NODE_ENV for now. I'm not sure there's enough use-cases to add to the Mix API.

Issue #2126, opened 2019-06-10
extract-text-webpack-plugin deprecation


JohnMica:
-   Laravel Mix Version:  "4.0.16"

### Description:

not sure if you are actually using this plugin for css or not but ```extract-text-webpack-plugin``` has been deprecated

https://github.com/webpack-contrib/extract-text-webpack-plugin

The recommended new plugin is https://github.com/webpack-contrib/mini-css-extract-plugin

### Steps To Reproduce:
just follow the link above and it seems that the plugin has been deprecated since the beta release

JeffreyWay:
Already known. We have to continue using extract-text-webpack-plugin until webpack 5 is for various compatibility reasons.

Issue #1954, opened 2019-01-22
Vendor css file and extractVueStyles, I'm missing something..


renzo-s:
-   Laravel Mix Version: 4.0.14
-   Node Version (`node -v`): 11.0.0
-   NPM Version (`npm -v`): 6.6.0
-   OS: OSX

### Description:
Hi all!

I'm trying to get rid of the injected styles in the `<head>` from node_module packages (vue components). I've been digging around several days now, but I'm guessing I'm missing a piece of the puzzle.

Tried several forms of;
`extractVueStyles: '[name].css'` => This never resulted in any extracted files.

After I couldn't get that to work, I found several similar questions / issues:

  - https://github.com/JeffreyWay/laravel-mix/issues/1589
  - https://github.com/JeffreyWay/laravel-mix/issues/1050
  - https://github.com/JeffreyWay/laravel-mix/issues/1896

Reading up on this, the `ExtractTextPlugin` / `mini-css-extract-plugin` seemed to make sense to try. The `ExtractTextPlugin` resulted in errors like `TypeError: Cannot read property 'find' of undefined`. Ok, this plugin is depricated for Webpack 4 anyway. Maybe there are unresolved issues.

Then `mini-css-extract-plugin` resulted in errors like `Error: Rule can only have one result source (provided use and loaders)`. I figured that Mix was already doing some extracting magic behind the scenes and the loaders were clashing.

So what am I missing to achieve creating a vendor.[HASH].css? 17 style tags (some of which are empty), with such a high specificity (since it's injected) is not really something to go into production with.

Thanks in advanced for any pointers.

My webpack.mix.js file:

```js

const mix = require('laravel-mix');
const webpack = require('webpack');

const domain = 'my-domain.mac'

mix.browserSync({
    host:  domain,
    proxy: 'https://'+domain,
    open: 'external',
    browser: 'google chrome canary',
    files: [
       /* .. not relevant for example .. */
    ],
    https: {
        key: `/Users/my-name/.valet/Certificates/${domain}.key`,
        cert: `/Users/my-name/.valet/Certificates/${domain}.crt`,
    }
})

mix
    .setPublicPath('html/')
    .disableNotifications()
    .webpackConfig({
        // Ignore all momentjs locales, decreases bundle size greatly
        plugins: [
            new webpack.IgnorePlugin(/^\.\/locale$/, /moment$/),
        ],
    })
    .options({
        // Tried several options like, true, vendor.css
[...]

PR #2132, opened 2019-06-14
Replace extract-text-webpack-plugin with mini-css-extract-plugin

andreyberezhnoy:
extract-text-webpack-plugin depracted, alpha version is not stable with webpack v4

This PR resolve some issues with CSS while using dynamic import or extractVueStyles

https://github.com/JeffreyWay/laravel-mix/issues/1954
https://github.com/JeffreyWay/laravel-mix/issues/2126

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -7188,8 +7188,7 @@
         "is-plain-obj": {
             "version": "1.1.0",
             "resolved": "https://registry.npmjs.org/is-plain-obj/-/is-plain-obj-1.1.0.tgz",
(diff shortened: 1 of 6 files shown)


andreyberezhnoy:
@JeffreyWay can you help me with tests? I don't understand what a problem is

QWp6t:
jway has been saying for a while that this can't be done until wp5. I don't know if he's expanded on exactly why tho.

laracasts:
We have to stick with that old plugin until webpack 5. Too many odd scenarios where compiles and CSS extraction breaks otherwise. 

If you can switch to mini-css-extract without the tests failing, I’ll merge it. Not sure that’s possible right now though.

PR #2160, opened 2019-07-16
Use collections and built-in over Lodash

ecrmnn:
- Prefer familiar and descriptive "Laravel collection syntax" with `collect.js`
- Remove Lodash for built-in standards like `isFunction`
- Removes unnecessary helpers


JeffreyWay:
Looks good! Thanks so much.

PR #2164, opened 2019-07-17
Adds ESLint

ecrmnn:
Hey @JeffreyWay 

I don't know if you want to add ESLint to this project, maybe not since you haven't already.
If you don't want it, just close it 🙂

I've only added rules that already fit your code style, except for `prefer-const` and `no-const-assign` that prefers `const` over `let`.

I've also added `eslint` under `scripts` in `package.json` and in `.travis.yml`.

- **prefer-const**
This rule is aimed at flagging variables that are declared using `let` keyword, but never reassigned after the initial assignment. If a variable is never reassigned, using the `const` declaration is better.

- **no-const-assign**
We cannot modify variables that are declared using `const` keyword. It will raise a runtime error.

- **newline-before-return**
his rule requires an empty line before `return` statements to increase code clarity, except when the `return` is alone inside a statement group (such as an if statement). In the latter case, the `return` statement does not need to be delineated by virtue of it being alone. Comments are ignored and do not count as empty lines.

- **semi**
This rule enforces consistent use of semicolons.

- **no-unreachable**
This rule disallows unreachable code after `return`, `throw`, `continue`, and `break` statements.

- **no-extra-semi**
This rule disallows unnecessary semicolons.

- **no-unexpected-multiline**
This rule disallows confusing multiline expressions where a newline looks like it is ending a statement, but is not.

- **comma-dangle**
This rule enforces consistent use of trailing commas in object and array literals.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,2 @@
+test/fixtures/fake-app/resources/assets/dynamic/dynamic.js
+test/fixtures/fake-app/resources/assets/extract/app.js 
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,24 @@
+{
+  "parserOptions": {
+    "ecmaVersion": 6,
+    "sourceType": "module"
+  },
+  "env": { 
+    "node": true
+  },
+  "rules": {
+    "no-const-assign": "error",
+    "newline-before-return": "error",
+    "semi":"error",
+    "no-unreachable": "error",
+    "no-extra-semi": "error",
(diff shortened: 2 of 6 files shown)


JeffreyWay:
Thanks, @ecrmnn. Happy to merge this in, but can you turn off `prefer-const`? I'm in the minority I know, but I prefer `let`.

ecrmnn:
Sure! I've turned it off.

Issue #2161, opened 2019-07-16
Notifications not showing in KDE 5.16 because of hardcoded timeout


AdrienPoupa:
-   Laravel Mix Version: 4.1.2 (`npm list --depth=0`)
-   Node Version (`node -v`): v12.6.0
-   NPM Version (`npm -v`): 6.10.1
-   OS: Kubuntu 19.04 with KDE 5.16

### Description:

I can see the notification for 2 ms instead of 2 seconds.

See screencast:

![mix](https://user-images.githubusercontent.com/15086425/61330156-08e7af80-a7ed-11e9-9a81-29cb4e4c1720.gif)

The problem is coming from Mix because if I use WebpackNotifierPlugin directly it works:

```
var WebpackNotifierPlugin = require('webpack-notifier');

mix.webpackConfig(webpack => {
    return {
        plugins: [
            new WebpackNotifierPlugin({alwaysNotify: true}),
        ]
    };
});
```

results in

![image](https://user-images.githubusercontent.com/15086425/61329591-ae018880-a7eb-11e9-8bf4-1868539c83e9.png)

After digging further, I understood that for KDE, setting a timeout to 2 is 2ms.

https://github.com/JeffreyWay/laravel-mix/blob/master/src/components/Notifications.js#L18

 If I set it to 2000 then I'm getting 2 seconds:
```
mix.webpackConfig(webpack => {
    return {
        plugins: [
            new WebpackNotifierPlugin({
                title: 'Laravel Mix 2',
                alwaysNotify: Config.notifications.onSuccess,
                hint:
                    process.platform === 'linux'
                        ? 'int:transient:1'
                        : undefined,
                timeout: 2000, // 2000 instead of 2
                contentImage: Mix.paths.root(
                    'node_modules/laravel-mix/icons/laravel.png'
                ),
            }),
        ]
    };
});
```

Interestingly if I set the title to Laravel Mix and not Laravel Mix 2, then the notification only shows for 2 ms. I guess this is because KDE suppresses duplicate notifications.

This PR is to blame: https://github.com/JeffreyWay/laravel-mix/pull/1890

Please revert it as it is not working universally, or make it configurable with no value by default.

KDE allows setting a timeout per program so it should not be done there IMO.

Thanks

AdrienPoupa:
It is not stale, fix is waiting to be merged

AdrienPoupa:
Still not stale

PR #2165, opened 2019-07-17
Make the notification timeout configurable and unset by default

AdrienPoupa:
KDE Plasma's notification timeout is set in milliseconds, not seconds.

This causes the hardcoded timeout of 2 to be interpreted as 2ms instead of 2 seconds and causes the notification to be unreadable.

Since the 2 is not a value working universally, I propose to make the timeout configurable and set it to undefined by default. On most desktop environments this seems to be 5 seconds. If this is too high for some users, they can either:
- Rely on their desktop environment to configure the timeout
- Set it themselves in Mix's config

Fixes https://github.com/JeffreyWay/laravel-mix/issues/2161

### DIFF
--- a/docs/os-notifications.md
+++ b/docs/os-notifications.md
@@ -7,3 +7,11 @@ mix.js(src, output).disableNotifications();
 ```
 
 Simple!
+
+You can set the timeout for the notifications with the following:
+
+```js
+mix.setNotificationsTimeout(2);
+```
+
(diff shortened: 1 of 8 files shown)


JeffreyWay:
I'd rather we remove the hardcoded timeout entirely. Making a new config option is overkill.

AdrienPoupa:
See https://github.com/JeffreyWay/laravel-mix/pull/2179

Issue #1979, opened 2019-02-06
PostCSS warning that it does nothing, no plugin is set


gabrielboliveira:
-   Laravel Mix Version: 4.0.14 (`npm list --depth=0`)
-   Node Version (`node -v`): v11.1.0
-   NPM Version (`npm -v`): 6.4.1
-   Yarn Version (`yarn -v`): 1.12.3
-   OS: macOS Mojave 10.14.2

### Description:

I have a pure Vue project with laravel-mix as bundler. Its a very simple repository, almost equal to [fireworkweb/vue-laravel-mix](https://github.com/fireworkweb/vue-laravel-mix). All css/scss are wrapped into `.vue` template files, so no call to mix.sass or any other css. 

After upgrading to v4, I keep getting this warning on every build (hot reload, dev and watch), but not when bundling for prod:

```
You did not set any plugins, parser, or stringifier. Right now, PostCSS does nothing. Pick plugins for your case on https://www.postcss.parts/ and use them in postcss.config.js.
```

You can check the [webpack.mix.js](https://github.com/fireworkweb/vue-laravel-mix/blob/master/webpack.mix.js) with how we use it.

### Steps To Reproduce:

Clone above repository and run `yarn dev`, `yarn hot` or even `yarn watch`.

luffyzhao:
have you solved it? i also met.

amcsi:
Have you tried doing this?

```js
mix
  .options({
    postCss: [
      require('autoprefixer'),
    ],
  })
```

luffyzhao:
> Have you tried doing this?
> 
> ```js
> mix
>   .options({
>     postCss: [
>       require('autoprefixer'),
>     ],
>   })
> ```
is ok!
thank you!

gabrielboliveira:
Thanks @amcsi, it does fixes the problem. Shouldn't mix add this by default?

amcsi:
I think so too...
@JeffreyWay ?

sebastiaanluca:
Having the same issue, seems like the PostCSS config array is not passed to Mix/Webpack.

```js
mix.options({
    autoprefixer: false,
    processCssUrls: false,

    postCss: [
        require('autoprefixer'),
        require('postcss-easy-import')(),
        require('tailwindcss'),
        require('postcss-nested')(),
        require('postcss-preset-env')(),
        require('postcss-css-variables')()
    ],
})
```

[...]

PR #2172, opened 2019-07-26
Use repo level postcss config if no custom config

LeoColomb:
Use `postcss.config.js` if the config file is present **and** `Mix.components.get('postCss').details[0].postCssPlugins` is empty.

May help to resolve: #1979


LeoColomb:
Hi @JeffreyWay, thanks for closing this PR. May I ask the reason why? Thanks again! 🙂

PR #2180, opened 2019-07-30
Update CssPurifier to laravel resources structure

Somethingideally:
Update CssPurifier to new laravel resources structure
resources/assets/js => resources/js


cmorbitzer:
This PR introduced a breaking change, and should be reverted to be released in a new major version. It's fine to update the plugin to automatically look for files in the new default locations, but we had a number of style rules disappear after a patch version increase and that shouldn't happen.

PR #2199, opened 2019-08-22
fix(hmr): using `css-hot-loader` instead of `style-loader`

u3u:
`css-hot-loader` can be better combined with `extract-text-webpack-plugin`
`style-loader` causes the style to be repeated (the styles of other pages are also bundled in)

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -4137,6 +4137,29 @@
                 "timsort": "^0.3.0"
             }
         },
+        "css-hot-loader": {
+            "version": "1.4.4",
+            "resolved": "https://registry.npmjs.org/css-hot-loader/-/css-hot-loader-1.4.4.tgz",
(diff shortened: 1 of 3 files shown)


u3u:
Before the merge, if someone wants to use it, it can be overridden with the `mix.override` method.

```js
mix.override((config) => {
  const sass = path.join(process.cwd(), '/resources/assets/sass');
  config.module.rules
    .filter(
      (item) => typeof item.test === 'string' && item.test.startsWith(sass)
    )
    .forEach((item) => {
      // You need to manually install the `css-hot-loader` module.
      item.use.splice(0, 1, 'css-hot-loader');
    });
});
```

thecrypticace:
This shouldn't be necessary anymore since we'll be using `mini-css-extract-plugin` which has builtin support for HMR.

cc @JeffreyWay

PR #2201, opened 2019-08-24
fix(sourcemaps): fix vue style source maps

u3u:
## API

- fix(api): change default source maps type to `inline-source-map`

## CSS

- fix(css): add `sourceMap` option for css loaders

## Vue

- fix(vue): replace `style-loader` to `vue-style-loader`


JeffreyWay:
Sorry for the delay, but I don't think any of this is needed anymore. Sourcemaps are applied automatically based on the devtool setting, which we already configured based on whether the user wants sourcemap support.

PR #2261, opened 2019-10-18
Run travis against node 12

bobbybouwmann:


bobbybouwmann:
@JeffreyWay It seems that the AppVeyor build is broken. Also the mac builds in travis don't seem to start. I'm not the only one with this issue if I look at some other PRs. 

Not sure if you can do anything against that, because it's a different party!

fitztrev:
#2248

PR #2279, opened 2019-11-15
Javascript component only for Javascript

pavinthan:
When I create new laravel project and build assets install `vue-template-compiler` dependency but It's not necessary if there are no Vue files, also It can be install when user use Vue presets so `mix.vue()` is good to apply Vue dependencies.

please advice me.


JeffreyWay:
It will be automatically installed until Mix 6 is out. You can safely ignore it for now.

pavinthan:
OK, thanks @JeffreyWay and is there any way to findout Mix 6 roadmap ?

Issue #1793, opened 2018-10-18
mix.sourceMaps() does not work


ryanmortier:
- Laravel Mix Version: 2.1.14
- Node Version (`node -v`): 10.12.0
- NPM Version (`npm -v`): 6.4.1
- OS: macOS

### Description:

Following the documentation here: https://laravel.com/docs/5.7/mix#css-source-maps

`mix.sourceMaps()` doesn't work. It's not adding source maps.

### Steps To Reproduce:

Add `mix.sourceMaps()` to your `webpack.mix.js` file and then `npm run dev`.

Here is my `webpack.mix.js` file:
```js
const mix = require('laravel-mix');

/*
 |--------------------------------------------------------------------------
 | Mix Asset Management
 |--------------------------------------------------------------------------
 |
 | Mix provides a clean, fluent API for defining some Webpack build steps
 | for your Laravel application. By default, we are compiling the Sass
 | file for the application as well as bundling up all the JS files.
 |
 */

mix.js('resources/js/app.js', 'public/js')
   .sass('resources/sass/app.scss', 'public/css');
```

SimplyThomas:
I ran into a similar problem earlier. #879 fix worked for me when I added the following to my options:
``` javascript
mix.options({
    processCssUrls: false,
})
   .sourceMaps()
   .js(...)
   .sass(...);
```
For me there was an [upstream issue](https://github.com/bholloway/resolve-url-loader/issues/28) with the `resolve-url-loader` dependency, but it recently got a re-write to postCSS and should resolve the issue. I don't have the time right now to work on a PR for this right now, however.

https://github.com/bholloway/resolve-url-loader/releases/tag/v3.0.0

ryanmortier:
Thanks, this workaround worked for me:

```js
if ( ! mix.inProduction()) {
    mix.webpackConfig({
        devtool: 'inline-source-map'
    })
}
```

I'm going to leave this issue open though as I believe the original option isn't working.

MDooley47:
Another workaround: `.sourceMaps(true, 'source-map')`. That first parameter is a boolean value for if you would source maps when ran as production as well. 

The interesting thing is that if you use the default value of `eval-source-map` everything seems to work fine. I'm not sure why it would work fine when you pass it in but not when you let it default to the value.

https://github.com/JeffreyWay/laravel-mix/blob/52582c120f2eaef0146d63a43c41235f40e3f030/src/Api.js#L12

[...]

PR #2282, opened 2019-11-20
Fix prod css/sass sourcemaps (#1793)

bentaber:
See Issue #1793

Sass sourcemaps fail to generate in production builds due to cssnano
combined with OptimizeCSSAssetsPlugin.  Moving cssnano into
postcss-loader resolves issue.

More detail https://www.webfoobar.com/node/109

Note eval-source-maps still will not work at all for CSS, so
implementers must use `Mix.sourceMaps(true, 'source-map')` to allow
development builds to successfully generate sourcemaps.

### DIFF
--- a/src/builder/webpack-default.js
+++ b/src/builder/webpack-default.js
@@ -39,14 +39,7 @@ module.exports = function() {
 
         optimization: Mix.inProduction()
             ? {
-                  minimizer: [
-                      new TerserPlugin(Config.terser),
-                      new OptimizeCSSAssetsPlugin({
-                          cssProcessorPluginOptions: {
-                              preset: ['default', Config.cssNano]
-                          }
-                      })
-                  ]
+                  minimizer: [new TerserPlugin(Config.terser)]
               }
             : {},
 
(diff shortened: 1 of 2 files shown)


bentaber:
Note all tests are passing when run locally. I notice last few commits do not pass appveyor ie: d432101c23198252e3952d8d89ee246d209265a8

JackWH:
I can confirm this PR solved the problem for me.

In my case, I was compiling LESS, SCSS, and JS files with Mix. I was trying to compile `app.js`, `app.less`, `error.less`, `landing.less`, `landing.js`, and `bootstrap.scss`.

After dumping the Webpack configuration to the console before Mix ran, it was mis-calculating the entry points as follows:

```
entry:       {
  '/js/app':     [
    '/Users/Me/Sites/MySite/resources/js/app/app.js',
    '/Users/Me/Sites/MySite/resources/less/app/app.less',
    '/Users/Me/Sites/MySite/resources/less/app/error.less',
    '/Users/Me/Sites/MySite/resources/less/landing/landing.less',
    '/Users/Me/Sites/MySite/resources/scss/bootstrap.scss',
  ],
  '/js/landing': ['/Users/Me/Sites/MySite/resources/js/landing/landing.js'],
}
```

This was causing only two sourcemaps to be generated, for `/js/app` and `/js/landing`, but none corresponding to the CSS files.

Manually patching this PR causes the entry points to calculate correctly — one for each file — and 6 corresponding sourcemaps to be generated, rather than just 2.

PR #2291, opened 2019-12-08
Update dependency to avoid security flawness

Thanerik:
It is an upgrade of a dependency to resolve the [CVE-2019-16769](https://github.com/advisories/GHSA-h9rv-jmmf-4pgx) security issue.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1455,6 +1455,22 @@
                 }
             }
         },
+        "aggregate-error": {
+            "version": "3.0.1",
+            "resolved": "https://registry.npmjs.org/aggregate-error/-/aggregate-error-3.0.1.tgz",
+            "integrity": "sha512-quoaXsZ9/BLNae5yiNoUz+Nhkwz83GhWwtYFglcjEQB2NDHCIpApbqXxIFnm4Pq/Nvhrsq5sYJFyohrrxnTGAA==",
+            "requires": {
+                "clean-stack": "^2.0.0",
+                "indent-string": "^4.0.0"
+            },
+            "dependencies": {
+                "indent-string": {
+                    "version": "4.0.0",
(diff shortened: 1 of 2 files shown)


checkerschaf:
What's the current status of it? I think this security issue should be addressed.

timrspratt:
The vulnerability was resolved with the patch update `2.1.1`, hence this change to bump to that version. Looking at the release log of the `serialize-javascript` package (https://github.com/yahoo/serialize-javascript/releases) it seems the bump from v1 to v2 has breaking changes. Have these been tested?

checkerschaf:
Not by me. But `serialize-javascript` is used by `terser-webpack-plugin` which is used by `webpack` which is then used by `laravel-mix`.
There are newer version of all packages in between but I'm not sure what breaking changes this comes with.

darrencoutts118:
@JeffreyWay @taylorotwell - could we have an update on this one? Its to patch a CVE.

Thanerik:
I don't know the status of this pull request. My trouble is that I don't know that much of Laravel Mix to fully understand the consquences of updating `serialize-javascript` to the newer version. I even don't know where the package is being used in. However I ran my laravel code with the updated package and nothing broke. But as I stated, I don't know the full usage of the package....

I can see that Webpack already has a pull request that would resolve this issue, but it's blocked: https://github.com/webpack/webpack/pull/10108

It's better if @JeffreyWay can take a look at this pull request, and said if we need to wait for Webpack to resolve it self?

darrencoutts118:
A little disappointing to see this has not progressed. For me, it looks like a reason to move from mix if security vulnerabilities don't get patched in a timely fashion.

timrspratt:
I think running `npm audit fix` will sort this without changes being needed. Let me know if not and we can get our pitchforks and head for @JeffreyWay 😂

PR #2303, opened 2020-01-04
[WIP] Webpack 5 Support

thecrypticace:
What it says on the tin. 

A few things:
1. extract-text-webpack-plugin cannot be used. It is not compatible with webpack 5 and it was deprecated even before webpack 4 was released. This will likely be the most difficult part of the conversion.
2. Not all loaders / plugins have been updated
3. PurifyCSS webpack plugin is no longer maintained and they suggest using the PurgeCSS plugin now.


JeffreyWay:
Going over this today. 👍

thecrypticace:
Awesome! I'll likely be working on this some either tonight or tomorrow.

thecrypticace:
Got quite a few more tests passing. CSS assets are now placed correctly and Vue styles should hopefully be (… if I can get them to compile).

Remaining problems I've noticed (so far — highly unlikely to be a complete list):
1. Async chunk splitting doesn't 100% honor the specified `webpackChunkName`. It generates something like `/vendors~js/split.js` instead of `/js/split.js`.
2. Vue is not compiling scss. From a cursory look it appears to be parsing scss as sass and I'm not sure why. Changing the sass options does not help afaict.
3. The combine script is leaking coffee script from an earlier script into the combined JS. (wat…)

thecrypticace:
I'm fairly certain that Vue Loader is causing some of the bugs here but testing that is rather complicated. It does not appear to be set up to be able to test multiple webpack versions.

I'll work on an isolated reproduction over there and see what I can do to resolve the problem.

andreyberezhnoy:
> 3\. The combine script is leaking coffee script from an earlier test into the combined JS. Running the test by itself passes. (wat…)

what about removing coffeescript support?

andreyberezhnoy:
> * Async chunk splitting doesn't 100% honor the specified `webpackChunkName`. It generates something like `/vendors~js/split.js` instead of `/js/split.js`.

@thecrypticace look at `automaticNamePrefix` and `automaticNameDelimiter` option
https://webpack.js.org/plugins/split-chunks-plugin/

thecrypticace:
Alright, I've done a good bit more work here. The tests mostly pass now.

1. There still a minor problem with Vue Loader but there's a workaround in place.
2. I still have to figure out style extraction into an existing CSS files. I believe this is the only failing test now.
3. Multiple CSS outputs are not working 100% properly at the moment. Fixing this should be rather simple I believe. (There's one test that had to change + a FIXME)

thecrypticace:
I've made some more progress. Vue style extraction mostly works. There is still a bit of weirdness there. There's been a regression w/ regards to async chunk splitting and multiple extractions. The styles are always being extracted now which is not ideal. (There's also an issue with filenames which I have to figure out)

We likely want to pull CSS into separate files only:
- When Vue requests it; or
- When a preprocessor is used

Rectifying these two goals is a bit complicated.

thecrypticace:
cc @JeffreyWay Okay I've fixed the weirdness and the tests are all passing now. 🎉 

This should now be in a much better place for you to look at.

JeffreyWay:
Okay, great! Thanks so much for all the work you've put into this. Going over it now.

thecrypticace:
@JeffreyWay 🎉 

I've tried this out on a fairly large existing app and found two things:
1. Module concatenation in webpack 5 has a bug with `delete`. If you use SweetAlert 2 this will pop up in a production build. I've already opened an issue w/ webpack regarding this.
2. There's still an issue related to CSS extraction preventing the page from loading even though the JS chunks are not generated but I can't quite figure that one out b/c in a vanilla (but very simplified) webpack config it works fine. I need to find a way to write an integration test for this but it seems like it'd be complicated unless we pull in something like playwright (which I can do if you like!). You will likely run into this one. I plan to attempt to track it down this weekend.

JeffreyWay:
Okay - going to make a list of things we need to address, as I test this out. Will keep updating this reply.

### Public Root

Looks like if I have this mix file:

```
mix
.js('resources/js/app.js', 'public/js')
.sass(
    'resources/sass/app.scss',
    'public/css'
);
```

...that will generate a `public/public/css/app.css` file. 

*Todo - Normalize the output path for CSS files.*

thecrypticace:
@JeffreyWay That's is what I was referring to yes. Same thing happens with `.postCss` as well (and likely the other pre-processors).

thecrypticace:
@JeffreyWay I had a bit of an aha moment and figured out what was causing it for postCss. The exclude rules for the `/\.css$/` test causes this to happen. **But I don't know the _why_ yet.** I have a feeling it is a similar situation for the other loaders but removing the exclude rules there causes other test failures (which I have yet to investigate).

If you comment out this block of code in `src/components/PostCss.js` the integration test passes:

```js
        config.module.rules.find(
            rule => rule.test.toString() === '/\\.css$/'
        ).exclude = this.details.map(postCss => postCss.src.path());
```

src/components/Preprocessor.js
```diff
@@ -149,6 +149,7 @@ class Preprocessor {
 
         src = new File(src);
 
+        output = File.stripPublicDir(output);
```
thecrypticace, line 152: I'mma be honest this feels super icky but seemed to be the only way for me to solve the problem. The File / path handling code seems to be somewhat messy. I wasn't sure how to work this into the general output path normalization. I'd be up for a larger refactoring of the file / path handling if you think it's warranted.

thecrypticace:
@JeffreyWay Okay, I believe I've addressed your 3 review points.

jasonlfunk:
I tried to use your branch on one of my rather complicated projects and came across two regressions:

1. I can no longer pass the `--env.mixfile` argument to change my webpack.mix file location. 

```
⬢ webpack: Unknown argument: --env.mixfile=resources/webpack.mix.js
```

2. The `globalVueStyles` setting seems to have no effect. It appears that the code still adds the sass-resource-loader, but never actually uses it.

Neither of these things appears to have tests, which is probably why they have fallen through the cracks.

thecrypticace:
@jasonlfunk thanks for testing! 💯 
I'll look at fixing these soon (today / tomorrow) and ensure tests exist for these.

src/components/Preprocessor.js
JeffreyWay: I agree, the file path handling in general is rough across the entire codebase. Let's address it later in a separate commit after we have webpack 5 support ready to go.

thecrypticace: 👍

JeffreyWay:
Okay, yeah just manually confirmed that those three issues I referenced seem to be fixed.

I've now moved on to testing this on a real codebase. I'll update this comment if/as I run into issues.

## Incorrect mix-manifest.json 

It looks like the `mix-manifest.json` file is not generating the correct file paths for CSS files. If I have:

```
mix.js('resources/js/main.js', 'public/js/app.js')
    .sass('resources/sass/app.sass', 'public/css');
```

That will produce the following `mix-manifest.json` file.

```
{
    "/js/app.css": "/js/app.css",
    "/js/app.js": "/js/app.js",
    "/js/manifest.js": "/js/manifest.js",
    "/js/vendor.js": "/js/vendor.js"
}
```

^ Notice the incorrect `/js/app.css` path that should be `/css/app.css`.

thecrypticace:
re: process not defined

webpack 5 removed automatic Node.js polyfills. Most likely this is a `process.env.SOMETHING_HERE` in the codebase that's not being replaced anymore. It's likely we'll need to augment uses of `webpack.DefinePlugin`.

thecrypticace:
@jasonlfunk

1. I've fixed globalVueStyles.
2. Webpack CLI complains about `env.anything` not being a known argument but it still works. ¯\_(ツ)_/¯

jasonlfunk:
Thanks for your updates. It has moved things along. I am running into this issue now: https://github.com/webpack-contrib/mini-css-extract-plugin/issues/487

Apparently `mini-css-extract-plugin` is not working with `hmr` and webpack-5 yet.

JeffreyWay:
@thecrypticace Yeah the `process` issue was related to an Algolia package I was using that references `process.env`. 

I'm going to merge this and tag the alpha release.

JeffreyWay:
We still have a ways to go, but...

```
npm install laravel-mix@v6.0.0-alpha.1
```

thecrypticace:
omg yus 🎉🎉🎉🎉

olach:
I just ran `npm i -D laravel-mix` and got this pre release alpha version (6.0.0-alpha.1) installed. Is this intended? I thought `npm install` would only install stable versions.

thecrypticace:
It was probably published with an incorrect distribution tag in npm.

cc @JeffreyWay see this article about this problem https://medium.com/@mbostock/prereleases-and-npm-e778fc5e2420

JeffreyWay:
Ooof, fixed.

PR #2324, opened 2020-02-10
Typescript type definitions

garbetjie:
I noticed that there is no Typescript types definition for Laravel Mix. I know that I find it useful to have code completion in my IDE for the libraries that I use (including Laravel Mix), and I suspect that there are others in the community that would find this helpful.

I've tried to find all the components that are loaded, and understand their arguments & return values. There is a possibility I may have missed one or two.

I'm still relatively new to Laravel Mix & Typescript, so any feedback on how to make this more useful for the community would be appreciated.


goldingdamien:
The types seem to have been created a while ago. What is the status of implementing this officially?

PR #2330, opened 2020-02-14
implement mix.when()

claytonrcarter:
Hi, this adds a small DX/QOL improvement: `mix.when()`. Mix has a wonderful, fluent API, and it often feels a little janky to me to break out of that when needing to do some conditional configuration. I hope that this helper will reduce that a little bit. This is of course inspired by Laravel's various `->when($condition, $callback)` helpers.

I added `mix.when()` to the stock `webpack.config.js`, but I wasn't sure if/where to add it to docs. If you could suggest the right place to add some docs, I'd be happy to do so. Oh, and this also includes a very minor docblock fix in `Api.js`.

# Example 

From one of my own `webpack.config.js`:
```js
mix.js('resources/js/app.js', 'public/js')
    .sass('resources/sass/app.scss', 'public/css')

if (process.env.NODE_ENV !== 'bundle') {
    mix.extract(['vue', 'vuex', 'axios'])
}

if (mix.inProduction()) {
    mix.version()
}
```

Which this PR, this could be rewritten as:
```js
mix.js('resources/js/app.js', 'public/js')
    .sass('resources/sass/app.scss', 'public/css')
    .when(
        process.env.NODE_ENV !== 'bundle',
        mix => mix.extract(['vue', 'vuex', 'axios'])
    )
    .when(
        mix.inProduction(),
        mix => mix.version()
    )
```

Thanks!


JeffreyWay:
Nice - thanks!

Issue #1891, opened 2018-12-26
Laravel Mix 4 installs vue-template-compiler package even if it is redundant


filipkrstic:
-   Laravel Mix Version: 4.0.12
-   Node Version: 8.11.1
-   NPM Version: 6.4.1
-   OS: Microsoft Windows 10

### Description:
I have started new Laravel project using latest version (5.7.19) and it is shipped with Laravel Mix 4.0.7. In this project I don't need any frontend scaffolding so I removed Vue scaffolding. After that when I ran `npm install` and then `npm run watch` I am getting this in the console:

> Additional dependencies must be installed. This will only take a moment.
>
>Running: yarn add vue-template-compiler --dev --production=false
>warning " > sass-loader@7.1.0" has unmet peer dependency "webpack@^3.0.0 || ^4.0.0".
>
>Okay, done. The following packages have been installed and saved to your package.json dependencies list:
>
>- vue-template-compiler

Why Laravel Mix 4 install this additional and obviously redundant package `vue-template-compiler` on the fly when I have removed Vue from scaffolding and how to get a rid of that package?

After all, I have tried to use Laravel Mix 4.0.7 and 4.0.12 - same thing in both cases.

### Steps To Reproduce:
1. Install new fresh copy of Laravel
2. Run `php artisan preset none`
3. Run `npm install`
4. Run `npm run watch`

vladshoob:
I have a similar problem without setting preset to none, but using default one shipped for Laravel.

I was able to run my npm run ... after
npm install  vue-template-compiler

JeffreyWay:
Mix has always installed this package. The only difference is it's now being pulled in locally. 

Mix was originally built with Vue in mind, so until the next major release, this dependency needs to stay. Just ignore it.

lkraav:
Duplicate of https://github.com/JeffreyWay/laravel-mix/issues/1877

mabasic:
I do not want to have vue-template-compiler dependency in my package.json file. Can it not be forced to install?

MunGell:
Hey, @JeffreyWay you mentioned that this dependency needs to stay until the next major release.
I am on v5 right at the moment and still getting it. Any chance removing it anytime soon?

paullaffitte:
If you need a quick-fix before my PR is merged, I also gived a hack to remove the dependency by hand. https://github.com/JeffreyWay/laravel-mix/issues/1877#issuecomment-596111404

Issue #1877, opened 2018-12-18
vue-template-compiler installed despite no reference to Vue


Log1x:
-   Laravel Mix Version: laravel-mix@4.0.12
-   Node Version (`node -v`): v11.4.0
-   NPM Version (`npm -v`): 6.4.1
-   OS: macOS 10.13.6

### Description:

I suspect starting in 4.0.10 with the introduction to [this](https://github.com/JeffreyWay/laravel-mix/commit/531ab623df61b13eeb711973b9bcc39594f6082b) commit, `vue-template-compiler` appears to be installing despite the project not using Vue nor having any reference to it whatsoever.

JeffreyWay:
`vue-template-compiler` has always been included with Mix. You're just seeing it more directly now. `mix.js()` does assume a Vue application, even if it doesn't require it.

billriess:
@JeffreyWay 

Why does `mix.react()` also assume a Vue application though?

mabasic:
I do not want to have vue-template-compiler dependency in my package.json file. Can it not be forced to install?

Kovah:
I would like to see an option to turn this off. In #1891 @JeffreyWay mentioned that Laravel Mix was build with Vue in mind but the current documentation indicates that it's more a universal build tool to make configuration with Webpack easier. Getting vue-template-compiler _force installed_ via Yarn even if I do not use any Vue and not using Yarn is irritating especially for starters.

If there is an option to turn this off or properly configure it, I would be pleased to get to know it. :)

othneildrew:
A **quick fix** that I found that was especially helpful for me was to use the `mix.scripts` method. This assumes an array of files to combine and does not install any vue dependencies. 

`webpack.mix.js` looks like this:
```js
mix.scripts(['resources/js/app.js'], 'public/js/app.js')
   .sass('resources/sass/app.scss', 'public/css');
```

If you're writing ES6+ JS like I am you can also use the `mix.babel()` which works actually the same but will use plugins like babel-loader to transform ES6+ code.

Hope this helps :smile:

paullaffitte:
Unfortunately, if you want to use the `import` keyword, you need modules and you need webpack, so you need to use the .js function.
I did a PR as mentionned above, but if it's not merge soon, you can use this **hack** (I checked the code of the library. If you're not using vue, it should not cause any issue, if you're using vue why are you reading this issue first?)

```js
const JavaScript = require('laravel-mix/src/components/JavaScript');

[...]

PR #2339, opened 2020-03-06
Remove vue-template-compiler when useless

paullaffitte:
Since several issues complains about vue-template-compiler being installed even when not necessary. I tried to reorganize a little bit how Vue and JavaScript components depends on each other in order to remove any dependencies on Vue when using `.js('source', 'dest')`.

Simply, instead of JavaScript depending on Vue, I made Vue depending on JavaScript, like it has already been done for Coffee, React, or similar. JavaScript should have the exact same behavior apart from anything related to Vue, and Vue should behave exactly the same. Components like Coffee and React should not be impacted since what is build with it should not relies on Vue.

Related issues:
- #1891
- #1877


JeffreyWay:
Sorry for just now getting to this.... 

The reason why the JavaScript component is linked to Vue is mostly due to legacy reasons. I've only brief looked at this commit, but it seems like all existing `mix.js()` calls that assume Vue compilation will have to be updated to `mix.vue()`, right?

If so - this of course is a major breaking change that will have to wait for the webpack 5 upgrade.

paullaffitte:
Yes, as you said it's actually a breaking change. Do you have any idea about when this update will be available? Is there anything I can do to help shipping this upgrade faster?
Otherwise I could also make some modification to keep this change as a minor one, like aliasing by default `mix.js()` on the Vue class and adding an option somewhere to alias `mix.js()` on Javascript until we do this major release?

sharifzadesina:
@JeffreyWay so we are not going to have this even in version 6.x?

paullaffitte:
I hope we will. I think JeffreyWay may have forgotten about it, but I don't know how should I contact him.

innocenzi:
This is the only change that I'm waiting for to no longer use Webpack Encore and switch back to Mix. Hope this PR doesn't get forgotten.

RomainMazB:
@JeffreyWay @paullaffitte 

Just to mention that the vue-template-compiler is only for Vue2, so this PR is may be mandatory for vue3 support.

Sources:
Evan You said:
> I think this is an issue that should be fixed in vue-cli-plugin-vue-next by not including vue-template-compiler (which is for 2.x only)
[vue-template-compiler throws with v3.0.0-alpha.4](https://github.com/vuejs/vue-next/issues/670#issuecomment-579785869)

And another issue and merged PR in vue-cli-plugin-vue-next to remove it for vue3 support in vue-cli-plugin:

[vue-template-compiler throws with v3.0.0-alpha.4](https://github.com/vuejs/vue-cli-plugin-vue-next/issues/5)

This maybe the first step for #2440

JeffreyWay:
We’ve split Vue from the core mix.js() to deal with this issue.

innocenzi:
Oh wow, it's happening. I'll be able to stop using Webpack Encore. Thanks. :D

paullaffitte:
So I guess you don't really enjoy when people contribute to your project, since you ended doing the same thing that I did again instead of just asking me to update my code. Kinda weird to call this "open source".

I think it's pretty obvious that I will not contribute to this project again.

JeffreyWay:
@paullaffitte No, it's a very different implementation that @thecrypticace worked on as part of the webpack 5 update.

PR #2358, opened 2020-03-31
Expose css-loader modules option

rafern:
VueJS Single File Components with SCSS modules don't work without the css-loader modules option enabled, so this PR exposes this option

### DIFF
--- a/src/components/Css.js
+++ b/src/components/Css.js
@@ -10,7 +10,7 @@ class Css extends AutomaticComponent {
                 test: /\.css$/,
                 loaders: [
                     'style-loader',
-                    { loader: 'css-loader', options: { importLoaders: 1 } },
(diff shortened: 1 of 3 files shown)


brianrlewis:
I just tested it and this pull it fixes my issue in #2370. Please merge or provide an alternate way to enable sass modules.

rafern:
Closing: laravel-mix is apparently being ported to Webpack5 and the code for the CSS component has changed, so I'm going to wait for the port to be complete and come back to it later in a new PR. You can still use the fork, I will only delete it when the devs finish porting

@brianrlewis
The last commit added CSS module auto-detection. VueJS <style module> tags are auto-detected (urls that have `module` in their query, auto-added by vue) and styles that have a name like `name.module.extension` e.g: `styles.module.css` or `styles.module.scss` also have module support enabled.

brianrlewis:
@rafern Awesome, module auto-detection is now working, thanks again.

Issue #2342, opened 2020-03-07
inProduction, mix.copyDirectory is not copying minified version of JS


rickertl:
-   Laravel Mix Version: 4.1.4
-   Node Version: 10.7.0
-   NPM Version: 6.14.2
-   OS: macOS 10.14.6

### Description:
In Production, I am using ‘mix.scripts' to combine 3 source JS files into 'js/all-scripts.js' and then using ‘copyDirectory' to duplicate this combined script at 'public/js’. Upon ‘npm run production', the 'js/all-scripts.js' is updated & minified (expected), but the ‘public/js/all-scripts.js’ is updated but NOT minified (unexpected).

I have worked around it by duplicating my ‘mix.scripts’ declaration and having the destination be ’public/js/all-scripts.js’, but seems I shouldn’t have to.

Am I missing something?

### Steps To Reproduce:
if (mix.inProduction()) {
mix.scripts([
    'src/js/script1.js',
    'src/js/script2.js',
    'src/js/script3.js'
], 'js/all-scripts.js’)
.copyDirectory('js’, ‘public/js’)
};

yneet:
@rickertl, do you have others files in directory js?

Have you test this code:
`mix.js([...], 'public/js/all-scripts.js')`

rickertl:
@yneet Yes, I have other scripts in the 'js' directory. I keep some separate that are only loaded for particular views.

I have tried mix.js, but it seems to fail with processing/combining bootstrap source. mix.scripts has no issue with that.

Thanks for the reply.

Chopinsky:
This problem sounds like a typical JavaScript async-problem, where minifiying a combined file requires disk IO, which could finish after the copy operation, and this would be a problem if the file-minification task and copy task are not using synchromized APIs. 

I'm not very familiar with this library, so if you could point me to the locations where the minification and copy tasks are generated, I can dig deeper on this issue.

rickertl:
@Chopinsky Your assessment sounds logical. That's what it seemed to me. Things were not happening in the correct order and if not dependent on one completing before the other (minify before copy), the odd result makes sense. 

Not sure where to point you. Perhaps someone else can help?

Thanks.

Chopinsky:
I think I spotted the issue: 

1. in `CustomTasksPlugin.js`, where build tasks are executed, we are looping over all scheduled tasks one after another. And for this particular issue, the file combination task and copy task will be executed consecutively at this point:

[...]

PR #2361, opened 2020-04-02
Fix issue #2342: minified files not copied in production

iastate-zuo:
This PR is to fix the issue as mentioned in #2342, where in production, the minified files are not copying to the output location in the first run. 

This fix removes the restriction where `CopyFileTask`s are skipped. To mitigate the risk for minifying a file multiple times, we will remember which files have already been minfied, and skip them when encountered a 2nd time. 

In addition, I've fixed a few `Vue` tests where the minified/uglified CSS contents show LF only differences.


JeffreyWay:
I think the main issue was that we weren't specifying which assets need to be minified after concatenating a directory of files. That should be fixed now.

PR #2368, opened 2020-04-03
Some fix regarding Vue dependencies

nmfzone:
- Javascript components: Only use Vue dependencies when `vue` is installed
- Vue components: Make sure `vue` added as dependencies


paullaffitte:
Looks like a duplicate of #2339, am I right?

JeffreyWay:
Sorry for the delay, but we’re handling this in a different way for upcoming release.

PR #2373, opened 2020-04-12
Fix yargs mixfile

marcwiest:
I'm using Mix outside of Laravel. It appears the original intent was to allow for a custom mix file location. However, I couldn't find a way to pass nested Yargs arguments to `argv`. It seams `argv.env.mixfile` does not work.

### DIFF
--- a/src/Paths.js
+++ b/src/Paths.js
@@ -28,7 +28,7 @@ class Paths {
      */
     mix() {
         return this.root(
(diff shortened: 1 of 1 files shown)


JeffreyWay:
Instead, we went with an environment variable to set the mix environment.

```
npx mix --mix-config="path/to/webpack.mix.js"
```

PR #2399, opened 2020-05-09
Load the PostCSS config file if no plugins have been specified

### LINKED ISSUE #2398: PostCSS config file is not used when all default plugins are disabled
-   Laravel Mix Version (`npm list --depth=0`): 5.0.4
-   Node Version (`node -v`): 10.20.0
-   NPM Version (`npm -v`): 6.14.4
-   OS: WSL @ Ubuntu 18.04

### Description:
PostCSS's config file `postcss.config.js` is not used when no plugins are specified, nor when autoprefixer is disabled. Introduced in v5 is cssnano which cannot be disabled. In v4 even when all plugins were disabled, the PostCSS config would not be loaded because the Preprocessor component always specified an empty array as the plugins key, so PostCSS wouldn't load its config file.

Fixed in #2399

### Steps To Reproduce:
- Create a fresh project with Laravel Mix
- Disable autoprefixer in Laravel Mix
- Create a postcss.config.js file and set the plugins you want (install their deps too, i.e. PurgeCSS)
- Observe that the PostCSS config file isn't attempted to be loaded / executed

glennmcewan:
Fixes #2398.

### DIFF
--- a/src/components/Preprocessor.js
+++ b/src/components/Preprocessor.js
@@ -37,45 +37,7 @@ class Preprocessor {
 
                     {
                         loader: 'postcss-loader',
-                        options: {
-                            sourceMap:
-                                preprocessor.type === 'sass' &&
-                                Config.processCssUrls
-                                    ? true
-                                    : Mix.isUsing('sourcemaps'),
-                            ident: `postcss${index}`,
-                            plugins: (function() {
-                                let plugins = Config.postCss;
-
-                                if (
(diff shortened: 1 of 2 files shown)


thecrypticace:
Hey @glennmcewan could you rebase this PR? If you don't have the time I could re-do this as well.

glennmcewan:
@JeffreyWay no problem, cheers!

PR #2460, opened 2020-08-10
Remove unused line of code

kanyinaf:
The first line in the shouldHotReload() method in Mix.js
creates a new file and then immediately deletes it. Removing
the line does not affect the tests.


thecrypticace:
`new File(…)` doesn't create a file. The `File` class is a collection of utilities for working with files. In this case it finds a file located at a given path and removes it if it exists.

Issue #2483, opened 2020-09-21
Extract all Vue logic from mix.js()


JeffreyWay:
As part of the big Mix 6 upgrade, we should take this opportunity to extract all Vue-specific logic from the `mix.js()` call. This will be a breaking change, but it's as good a time as any to do it. Like `mix.react()`, we'll need to add a `mix.vue()` command.

thecrypticace:
I'm working on #2481 / #2482 / #2483 (this issue) right now. How are you thinking this will work with things like `mix.ts()`? Right now they hardcode support for Vue (but not react). For me it seems like `mix.js` should still be used but `mix.vue()` and `mix.react()` are effectively _features_ that can be enabled for use in any used language.

My thinking is we could write something like the following:

```js
// 1. enable Vue support
mix.vue()

// 2.a. Compile app JS
mix.js("app/js/app.js", "public/js/app.js")

// 2.b. alternatively compile app TypeScript
mix.js("app/ts/app.ts", "public/js/app.js")
```

JeffreyWay:
I'll talk with you in the PR thread.

PR #2484, opened 2020-09-22
[WIP] Add mix.vue

thecrypticace:
First attempt supporting Vue 2 and Vue 3 as well as introducing the new mix.vue API mentioned in #2483.

This is a fairly big work in progress and I'm absolutely looking for feedback. I left a comment w/ an idea on how I think `mix.vue()` (as well as `mix.react()`) should work but I'll mirror it here to help centralize discussion:

It seems like `mix.js(…)` should still be used but `mix.vue()` and `mix.react()` are effectively _features_ that can be enabled for use in any language supported by mix.

I have not implemented this approach (yet) but my thinking is we could write something like the following:

```js
// 1.a. enable Vue 2.x support
mix.vue2() // or mix.vue3()

// 1.b. Alternative API which helps mentally enforce the idea that only one or the other is supported at one time. Not both.
mix.vue({ version: 2 }) // or mix.vue({ version: 3 })

// 2.a. Compile app JS
mix.js("app/js/app.js", "public/js/app.js")

// 2.b. alternatively compile app TypeScript
mix.ts("app/ts/app.ts", "public/js/app.js")
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,9 +96,9 @@
         "sinon": "^9.0.1",
         "stylus": "^0.54.5",
         "stylus-loader": "^3.0.2",
-        "vue": "^2.5.21",
+        "vue2": "npm:vue@^2.6.12",
         "vue-style-loader": "^4.0.2",
-        "vue-template-compiler": "^2.5.21"
+        "vue-template-compiler": "^2.6.12"
     },
     "engines": {
         "node": ">=12.14.0"
--- a/package.json
+++ b/package.json
@@ -73,6 +73,7 @@
     },
     "devDependencies": {
         "@babel/preset-react": "^7.0.0",
+        "@vue/compiler-sfc": "^3.0.0",
         "ava": "^1.4.1",
         "coffee-loader": "^1.0.0",
         "coffeescript": "^2.2.3",
@@ -97,6 +98,7 @@
         "stylus": "^0.54.5",
         "stylus-loader": "^3.0.2",
         "vue2": "npm:vue@^2.6.12",
+        "vue3": "npm:vue@^3.0.0",
         "vue-style-loader": "^4.0.2",
         "vue-template-compiler": "^2.6.12"
     },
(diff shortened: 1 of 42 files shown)


JeffreyWay:
Thanks for getting started on this. 

Hmm, I don't love `mix.vue2()` and `mix.vue3()`. Do you think we can guess which Vue component class to use based on the version of Vue that the user currently has installed? That way, we can stick with `mix.vue()` and it'll just magically work.

thecrypticace:
> Hmm, I don't love mix.vue2() and mix.vue3()

I 100% agree. I was thinking about it this morning and was going to see what I could do to auto-detect the installed Vue version because vue2/vue3 feels weird now that I've had some time to think on it.

> Yeah, this sounds reasonable. Do you mean that it would be like this...

Exactly. And since mix calls are chainable these will all work:
```
// As chained method calls
mix.js().vue({ version: 2 });
mix.vue({ version: 2 }).js();

// As separate method calls (order should not matter)
mix.js();
mix.vue({ version: 2 });
```

JeffreyWay:
Great. Sounds good to me. It's definitely a breaking change, but dealing with this has been on my list for a couple years now. I'd been avoiding it, but now seems like the best time to finally do it.

thecrypticace:
I've prototyped a version of `mix.vue()`. Let me know what you think of the implementation.

There are still a few points I want to address (like centralizing the style handling for Vue extraction _if_ possible so no component is checking `Config.extractVueStyles` and `Config.globalVueStyles`).

JeffreyWay:
Okay, this is looking pretty good to me. Treating `mix.vue()` as a feature opt-in is pretty smart, because it now gives us a place to put all those toggles that were in the Config file. 

Let's get this merged in, and then one of us can migrate all Vue-specific config options to the component.

thecrypticace:
Alright I pushed a commit to excise Vue from everything (except for ts-loader's `appendTsSuffixTo` option which I'm not sure how to put in just the Vue component instead of the typescript one but I think it's probably okay to leave it as is).

PR #2488, opened 2020-09-25
Revert "Respect processCssUrls option for Css-component"

JeffreyWay:
Reverts JeffreyWay/laravel-mix#2433


davidwebca:
@JeffreyWay Just ran into an issue where I noticed processCssUrls wasn't respected. Any details on why this revert commit?

PR #2491, opened 2020-09-26
CLI tool improvements

thecrypticace:
Still working on this but figured I'd open the PR so you could get a look at it

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -2,39 +2,78 @@
 
 const { Command } = require('commander');
 const { spawn } = require('child_process');
+const debug = require('debug')('laravel-mix');
 
-const program = createProgram();
+run();
 
-executeScript();
+async function run() {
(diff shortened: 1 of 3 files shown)


bin/cli.js
```diff
@@ -2,39 +2,84 @@
...
     program.version('0.0.1');
+    program.option(
+        '--mix-config <path>',
+        'Specify mix config file',
+        mixConfigPath
+    );
+    program.option(
+        '--webpack-config <path>',
+        'Specify webpack config file',
+        webpackConfigPath
```
JeffreyWay, line 28: So what's the usecase for allowing a user to override the webpack config file that Mix uses?

thecrypticace: Thinking about it… it shouldn't ever need to change. Not even for testing so I've removed this option.

PR #2493, opened 2020-09-26
Mix 6 typescript typings

thecrypticace:
Supersedes #2324

Thanks for the head start @garbetjie it was super helpful!


JeffreyWay:
Okay, I'm a dummy when it comes to TypeScript. What's the workflow for keeping this file up to date? Do we have to manually update every time? Is there a tool that will do this automatically?

thecrypticace:
Unless mix itself was written in typescript I'm not sure that there is a way to keep this automatically updated.

PR #2502, opened 2020-10-02
Customize dynamic import paths when compiling JS

### LINKED ISSUE #2500: Dynamic Import Output Path
In Mix 6, dynamic imports seem to be outputting to the public directory.

Screen Shot 2020-10-01 at 3 58 45 PM

We can of course override this, like so:

```js
import('./module' /* webpackChunkName: "js/module" */).then();
```

But I wonder if we should instead detect where JavaScript assets are being compiled to, and use that path as the default.

thecrypticace:
Fixes #2500 

It doesn't work for TypeScript yet. Got some things to set up before I can properly test it.

### DIFF
--- a/test/fixtures/fake-app/resources/assets/extract/app.js
+++ b/test/fixtures/fake-app/resources/assets/extract/app.js
@@ -4,6 +4,6 @@ import 'core-js';
 new Vue({
     components: {
         VueSplit: () =>
-            import(/* webpackChunkName: 'js/split' */ './VueSplit.vue')
+            import(/* webpackChunkName: '/js/split' */ './VueSplit.vue')
     }
 }).$mount('#app');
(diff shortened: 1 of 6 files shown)


thecrypticace:
Oof there's a test failure w/ path concatenation in some cases. Will fix.

test/fixtures/fake-app/resources/assets/extract/app.js
```diff
@@ -4,6 +4,6 @@ import 'core-js';
 new Vue({
     components: {
         VueSplit: () =>
-            import(/* webpackChunkName: 'js/split' */ './VueSplit.vue')
```
thecrypticace, line 7: @JeffreyWay  How do you want to handle this? I opted in the new code to check for absolute paths and not touch it if there is one but should we also detect the relative path if it starts with the same name?
Seems like it could be brittle (but still doable).

JeffreyWay: Hmm, will need to think about that. For now, my instinct is no. Will come back to this after I play around with it a bit.

JeffreyWay:
This PR does break some of the tests, unfortunately. Looks like in certain cases, the output directory gets doubled. 

Screen Shot 2020-10-02 at 11 47 27 AM

PR #2508, opened 2020-10-03
Add support for watch polling

thecrypticace:
Adds a `--poll` option to the mix CLI. Also fixes the test suite and disables the output of the friendly errors plugin when running tests so the console appears less spammy.


JeffreyWay:
We could do this, but I wonder if we should just setup a pass-through option that pipes everything to the webpack script. That way, you still have full access to every webpack option.

thecrypticace:
Agreed. Now you can do `mix watch -- --watch-poll`

PR #2510, opened 2020-10-04
Revert "Customize dynamic import paths when compiling JS"

JeffreyWay:
Reverting this temporarily due to the tests it broke.


thecrypticace:
The broken test was a missed import change in `PureSplit.vue` which I fixed while I was in the watch polling PR.

JeffreyWay:
Oh okay. I'll have a look and revert the revert if we're good to go. 👍

PR #2511, opened 2020-10-04
Better stats output

thecrypticace:
Work in progress. Should be a good starting point. I've added the branch to the mix repo so we can both easily iterate on this as needed until we're good with it.


JeffreyWay:
Nice work. Thanks for handling this.

PR #2525, opened 2020-10-07
Enable better side effect optimization

thecrypticace:
Upgrades webpack & mini-css-extract-plugin for transitive side-effect detection and enables side effect optimization in production.


JeffreyWay:
Ugh, this crap is so confusing haha.

thecrypticace:
😂

Issue #2518, opened 2020-10-05
[6.x] Can't resolve 'postcss-loader'


mbardelmeijer:
-   Laravel Mix Version: 6.0.0-beta.4 (`npm list --depth=0`)
-   Node Version (`node -v`): v12.18.3
-   NPM Version (`npm -v`): 6.14.6
-   OS: WSL2 on Windows

### Description:
When trying to compile Tailwind CSS with the postcss-loader, it gives an error. It does seem to work under the 5.x branch. 

### Steps To Reproduce:
Setup your `webpack.mix.js` like:
```
const mix = require('laravel-mix');

mix.js('resources/js/app.js', 'public/js')
    .postCss('resources/css/app.css', 'public/css', [
        require('postcss-import'),
        require('tailwindcss'),
    ])
    .version();
```

`package.json`
```
{
    "private": true,
    "scripts": {
        "production": "cross-env NODE_ENV=production node_modules/webpack/bin/webpack.js --config=node_modules/laravel-mix/setup/webpack.config.js"
    },
    "devDependencies": {
        "cross-env": "^7.0",
        "laravel-mix": "^6.0.0-beta.4",
        "resolve-url-loader": "^3.1.0"
    },
    "dependencies": {
        "@tailwindcss/ui": "^0.6.2",
        "postcss-import": "^12.0.1",
        "tailwindcss": "^1.8.10"
    }
}
```

Run `npm run production`, the following error is returned: `Module not found: Error: Can't resolve 'postcss-loader' in 'X'` is returned.

mbardelmeijer:
Welp did work after running `yarn add postcss-loader` ofc. Nvm 🙈

Issue #2517, opened 2020-10-05
[6.*] output path for PostCSS seems to be buggy


mohamadnorouzi20:
-   Laravel Mix Version: 6.0.0-beta.4
-   Node Version: v12.18.3
-   NPM Version: 6.14.8
-   OS: Fedora 32

### Description:
This used to work just fine in Mix 5 (pay attention to `build` directory and also the `.pcss` extension)
```js
const mix = require('laravel-mix')

mix.js('resources/js/app.js', 'public/build/js')
mix.postCss('resources/pcss/app.pcss', 'public/build/css', [
    //
])
```

But now I'm getting this after running `npx mix build`:
![image](https://user-images.githubusercontent.com/8435784/95107982-7e6bd280-0747-11eb-8e75-0ffe6b0ff316.png)

It's dumping the `app.css` file inside the `js` directory

mohamadnorouzi20:
I checked a few things and it seems this only happens because of the `.pcss` extension. it seems like if the extension does not belong to some list of allowed extensions it will ignore the **postCss output path** and use **js output path** instead.

JeffreyWay:
I don't think we've ever supported the `pcss` extension, so I'm confused how it was working in Mix 5.

thecrypticace:
I _think_ this should still place the file in the right location regardless of extension. I'll check on that.

cds-mm:
Experiencing a similar issue using `.sass` except it is producing 2 `.css` files in different directories:
```javascript
const mix = require('laravel-mix')

mix.js('src/app.js', 'public/js').vue({version: 2})
mix.sass('src/app.scss', 'public/css')
```
![image](https://user-images.githubusercontent.com/48589485/95230896-6012df00-07fa-11eb-8887-02a4f6e71790.png)

thecrypticace:
@cds-mm The only way I can get multiple output files like that is to run postCss + sass for separate files with the same output file name. Can you post a minimal reproduction?

thecrypticace:
@mohamadnorouzi20 @JeffreyWay We don't support `.pcss` extensions. We could though. All it would require is changing the `chunkRegex` for postCSS.

JeffreyWay:
This issue might be related: #2522

Something odd does seem to be happening with CSS extraction is certain cases.

cds-mm:
[...]

PR #2526, opened 2020-10-07
Fix compilation with multiple and imported css assets

thecrypticace:
An attempt at fixing #2517 and #2518

🤞


thecrypticace:
Real smart one I am forgetting to commit the file mix is actually compiling. 😂

PR #2530, opened 2020-10-08
Extract only used chunks + configurable extraction support

### LINKED ISSUE #2507: mix.extract(["lib1", "lib2"]) duplicates instead of splits even when libraries are used
This is kinda a backwards compatability thing with the way it works right now but it's something that should be fixed because it's definitely counterintuitive because it's not extracting dependencies — it's separately compiling those libraries into another chunk — or referencing them from vendor.js if you have a call to `mix.extract()` without any arguments.

This means that code like this (taken from a real project I'm refactoring)…
```js
mix.extract()
mix.extract([
  "codemirror", "easymde", "marked", "dompurify",
], "js/vendor~markdown.js")
mix.extract([
  "billboard.js", "d3", "d3-transition", "d3-selection", "d3-scale", "d3-time-format", "d3-brush", "d3-zoom", "d3-color", "d3-axis"
], "js/vendor~charts.js")
mix.extract([
  "vue", "portal-vue", "v-calendar", "v-tooltip", "@vue/composition-api"
], "js/vendor~vue.js")
```

… does not work as expected.

My thoughts are to improve the behavior of mix.extract and provide an enhanced API to support the necessary level of customization (and allow for further customization in the future without a drastic change in the API or syntax)

General idea

```js
// Extract everything in node_modules
mix.extract()

// Compile lib1 and lib2 into a separate file
// Even if some are not used
// I think this behavior should be changed since we're making some backwards-incompatible changes
// But we'll still need an escape hatch for people who need the old behavior
mix.extract(["lib1", "lib2"])

// Extract lib1 & lib2
// Only if used
mix.extract({
  libraries: ["lib1", "lib2"],
  // when: 'used' is the default
})

// Extract lib1 & lib2
// Always extract into a separate entry file
// For now when: always would only support the array variant of libraries
mix.extract({
  libraries: ["lib1", "lib2"],
  when: 'always',
})

// Extract any used d3-related modules
mix.extract({
  libraries: /d3|d3-.*/
})

// Completely customize the test
// Can be a function, regex, or other expected by webpack
mix.extract({
  test: /lib1|lib2/
})
```

thecrypticace:
Closes #2507

I've opted to change the behavior to never use entries for vendor extraction per the webpack documentation (https://webpack.js.org/concepts/entry-points/). If the user wants this behavior back I believe they'd need to create a file that imports all of the necessary libraries and compile it. Because of this behavior change I've not included the `when` parameter.

### DIFF
--- a/src/builder/Entry.js
+++ b/src/builder/Entry.js
@@ -50,39 +50,6 @@ class Entry {
     }
 
     /**
-     * Add a vendor extraction.
-     *
-     * @param {Object} extraction
-     */
-    addExtraction(extraction) {
-        if (!Mix.bundlingJavaScript && !extraction.output) {
-            throw new Error(
-                'Please provide an output path as the second argument to mix.extract().'
-            );
-        }
-
-        let outputPath = extraction.output;
-
-        // If the extraction output path begins with a slash (forward or backward)
-        // then the path from the public directory determined below will be wrong
-        // on Windows, as a drive letter will be incorrectly prepended to
-        // (e.g. '/dist/vendor' -> 'C:\dist\vendor').
-        let startsWithSlash = ['\\', '/'].indexOf(outputPath[0]) >= 0;
-        outputPath = startsWithSlash ? outputPath.substr(1) : outputPath;
-
-        let vendorPath = outputPath
-            ? new File(outputPath)
-                  .pathFromPublic(Config.publicPath)
-                  .replace(/\.js$/, '')
-                  .replace(/\\/g, '/')
-            : path.join(this.base, 'vendor').replace(/\\/g, '/');
-
-        this.add(vendorPath, extraction.libs);
-
-        return vendorPath;
-    }
-
-    /**
      * Add a default entry script to the structure.
      */
     addDefault() {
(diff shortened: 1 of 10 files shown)


JeffreyWay:
Okay, thanks. I'll need to take a little time to read and learn what you've done here. If this is a breaking change, can you also add a quick commit here that documents it in the upgrade.md file? If there's a before/after example you can provide, please include.

thecrypticace:
There is a breaking change when using `mix.extract(["library1", "library2"])`. I'll add some notes to the upgrade guide.

thecrypticace:
Two things:
1. I'm still tweaking the typescript defs here. Trying to figure out why VSCode is yelling…

2. I think I need to add priority selection to make this work:
```js
mix.extract([
  "vue", "portal-vue", "v-calendar", "v-tooltip", "@vue/composition-api"
], "js/vendor~vue.js")
mix.extract()
```

It's be preferable if the specified libraries had a higher extraction priority so we still end up with multiple bundles.

thecrypticace:
Alright as long as the tests pass this should be good to go

PR #2534, opened 2020-10-09
Fix file paths for dynamic imports

thecrypticace:
Fixes dynamic imports having nested public path folders

cc @ninjaparade


ninjaparade:
@thecrypticace 🙌🙌🙌

Issue #1972, opened 2019-02-01
manifest path is wrong


dvlpr91:
Hi...

my **webpack.mix.js**
```
...

mix.extract(['axios'], 'public/js/core').autoload('axios');

mix.js([
  'resources/js/a/app.js'
], 'public/js/a');

mix.js([
  'resources/js/b/app.js'
], 'public/js/b');

...

mix.version();
```

**result**

...
public/js/core.js (vendor)
public/js/b/manifest.js
...

what?

msonowal:
@dvlpr91 
Did you manage to fix this?

dvlpr91:
> @ dvlpr91
> 이 문제를 해결할 수 있었 습니까?

No, I cant.
This can not be done.

dvlpr91:
> @dvlpr91
> Did you manage to fix this?

No, I cant.
This can not be done.

kebien6020:
After looking around in the source code I managed with this:

```js
mix .override(config => {
        config.optimization.runtimeChunk.name = '/js/manifest'
    })
```

adampatterson:
This was closed, but I have the same issue only it works with the `.js` but the `.react` seems to be causing the issue now.

Issue #2486, opened 2020-09-23
vendor.js exported in the wrong location.


adampatterson:
-   Laravel Mix Version: 5.0.5
-   Node Version (`node -v`): 10.16.0
-   NPM Version (`npm -v`): 6.14.8
-   OS: macOS 0.15.6

### Description:
I have two scripts, one exports to a subfolder and that is where the `vendor.js` and `manifest.js` file end up.

I also tried a few other scenarios, I don't think I am misunderstanding anything. I believe the vendor and manifest files should be placed in root of my directory. 

### Steps To Reproduce:

webpack.mix.js

```
const mix = require('laravel-mix')

const extractLibs = [
    'react',
    'react-dom',
    'lodash',
    'jquery',
    'bootstrap'
]

mix.react('resources/js/wizard/wizard.js', 'public/js/wizard')
   .js('resources/js/app.js', 'public/js')
   .sass('resources/sass/app.scss', 'public/css')

mix.extract(extractLibs)

mix.version()
```

**wizard.js**
```
import React from 'react'
import ReactDOM from 'react-dom'

// React stuff
```

**app.js**
```
import $ from 'jquery'
import 'bootstrap'

import 'slick-carousel'

// Code here
```

The output looks like this.
```
/js/app.js
/js/wizard/manifest.js <-- I want this to be in the JS root.
/js/wizard/vendor.js <-- I want this to be in the JS root.
/js/wizard/wizard.js
```

If I remove the React script both the `vendor.js` and `manifest.js` are in the correct path, if I don't export the React script to a subfolder then both the `vendor.js` and `manifest.js` are in the correct path.

if I create another react entry and export the file to the root JS folder then that is where the `vendor.js` and `manifest.js` files are exported as well.

```
mix.react('resources/js/wizard/wizard.js', 'public/js/wizard')
   .react('resources/js/wizard/wizard2.js', 'public/js/') // vendor and manifest will end up in /js
   .js('resources/js/app.js', 'public/js')
   .sass('resources/sass/app.scss', 'public/css')
```

This example will place the `vendor.js` and `manifest.js` files under `/wizard`
```
mix.react('resources/js/wizard/wizard.js', 'public/js')
   .react('resources/js/wizard/wizard2.js', 'public/js/wizard') // vendor and manifest will end up in /wizard
   .js('resources/js/app.js', 'public/js') // Does not matter where this script is placed, it does not impact the location of the vendor and manifest file
[...]

PR #2542, opened 2020-10-12
Adds an ability to specify the path for the `manifest.js`

hivokas:
`mix.extract()` creates `vendor.js` & `manifest.js`. While that's possible to specify the directory for the `vendor.js`, that's not possible to do the same for `manifest.js`.

The destination directory for the `manifest.js` is detected automatically. And the current behavior is quite unobvious.

```
mix
  .extract()
  .js('resources/js/app.js', 'public/js')
  .js('resources/js/forms/signup.js', 'public/js/forms');

// `manifest.js` will be stored to the `public/js/forms`
```

```
mix
  .js('resources/js/app.js', 'public/js')
  .extract()
  .js('resources/js/forms/signup.js', 'public/js/forms');

// `manifest.js` will be stored to the `public/js/forms`
```

```
mix
  .js('resources/js/forms/signup.js', 'public/js/forms')
  .extract()
  .js('resources/js/app.js', 'public/js');

// `manifest.js` will be stored to the `public/js`
```

This PR adds an ability to specify the directory for the `manifest.js` explicitly the following way:

```
mix.options({ manifestJsPath: 'js' });
```

Fixes https://github.com/JeffreyWay/laravel-mix/issues/1972 https://github.com/JeffreyWay/laravel-mix/issues/2486

### DIFF
--- a/docs/extract.md
+++ b/docs/extract.md
@@ -38,6 +38,16 @@ Once you compile your code - `npx mix` - you'll find three new files: `app.js`, 
 
 While it's true that we're now importing three scripts instead of one, the benefit is improved long-term caching of vendor code that rarely changes. Further, HTTP2 makes the cost of importing multiple scripts a non-issue.
 
+### Customizing the runtime chunk (`manifest.js`) path
+
+By default, the runtime chunk (`manifest.js`) is generated to a public directory.
+
+However, the path can easily be customized:
+
+```js
+mix.options({ runtimeChunkPath: 'js' });
+```
+
 ### The Manifest File
 
 You might still be confused by that third `manifest.js` file. Webpack compiles with a small bit of run-time code to assist with its job. When not using `mix.extract()`, this code is invisible to you and lives inside your bundle file. However, if we need to split our code, that runtime code must "live" somewhere. As such, Laravel Mix will extract it to its own file.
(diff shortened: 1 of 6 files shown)


JeffreyWay:
Rather than making this configurable, should we instead place it wherever the `vendor.js` is saved?

hivokas:
@JeffreyWay do I understand correctly that currently it’s possible to make multiple extractions?

thecrypticace:
It is possible to make multiple vendor extractions so something like this seems reasonable. I _think_ I'd name it either `manifestPath` or `runtimePath` as that's what `manifest.js` is — the webpack chunk loading runtime.

hivokas:
> I _think_ I'd name it either `manifestPath` or `runtimePath` as that's what `manifest.js` is — the webpack chunk loading runtime.

Thanks for your suggestions. I've decided to rename `manifestJsPath` to `runtimeChunkPath` because the webpack uses this term internally (https://webpack.js.org/configuration/optimization/#optimizationruntimechunk).

> Rather than making this configurable, should we instead place it wherever the `vendor.js` is saved?

As stated above, that's possible to have multiple extractions, so I think that having an ability to explicitly specify the runtime chunk path is the most flexible option.

hivokas:
I've provided the test and updated the documentation & upgrade guide.

JeffreyWay:
Nice work, @hivokas. Thanks so much for taking care of this. 👍

PR #2544, opened 2020-10-13
Fallback to entry base for runtime chunk path

thecrypticace:
I missed this during review of the PR. We only want to change the path of `manifest.js` if the user sets the config option. Setting it to empty string places it outside of the JS directory by default which we don't want.


hivokas:
@thecrypticace please update the upgrade guide as well

thecrypticace:
Ah missed that. Updating…

PR #2575, opened 2020-10-22
Install PostCSS 8.x when running Mix for the first time

thecrypticace:
This installs PostCSS 8 into the user's root package.json after first run of Mix. This should, hopefully, squash most / all those weird PostCSS problems we're having.


Vinlock:
Thank you for this!

PR #2601, opened 2020-11-06
Fix sourcemap config type

### LINKED ISSUE #2600: Sourcemaps don't work according to types or docblock
-   Laravel Mix Version: `6.0.0-beta.11`
-   Node Version (`node -v`): `v14.15.0`
-   NPM Version (`npm -v`): `6.14.8`
-   Yarn version (`yarn -v`): `1.22.5` 
-   OS: `Ubuntu 20.04.1 LTS`

### Description:

I noticed some weird behaviour surrounding source maps.

First off, the Typescript (and therefore IDE-autocomplete) wrongly suggests `sourceMaps`, while `sourcemaps` is expected in the config. It also specifies `sourceMaps` to be a boolean, and the same mistake is in `src/config.js`, while  `false`  is [a valid devtool][1] but `true` is not. 

This also means a simple `sourcemaps: !mix.inProduction()` won't work, but would require a ternary expression.

Maybe I'm doing this entirely wrong (is there a `mix.sourcemaps` API?), but it should be fixed nevertheless.

As a list:

- `src/config.js` and the actual code uses `sourcemaps`
- `sourcemaps` should be false or [a valid devtool][1].
- `typings/index.d.ts` contains `sourceMaps?: boolean`, should be `sourcemaps?: (false|string)` (or something, I don't know Typescript).
- Maybe `sourcemaps: true` should evaluate to using the `source-map` devtool.

[1]: https://webpack.js.org/configuration/devtool/#root

### Steps To Reproduce:

1. Install `laravel-mix@next`

#### Expected behaviour with IDE-autocomplete (VS Code):

2. Make a `webpack.mix.js`:
```js
const mix = require('laravel-mix')
// …
mix.options({
  sourceMaps: !mix.inProduction()
})
```

##### Expected result:

- on dev build: source maps are created
- on prod build: no source maps are created.

##### Actual result

No source maps are ever created.

#### Expected behaviour using `src/config.js` as leading

2. Make a `webpack.mix.js`:
```js
const mix = require('laravel-mix')
// …
mix.options({
  sourcemaps: !mix.inProduction()
})
```

##### Expected result:

- on dev build: source maps are created
- on prod build: no source maps are created.

##### Actual result

- on dev build: build fails.
- on prod build: no source maps are created.

### Workaround

There is, of course, a simple workaround:

```js
const mix = require('laravel-mix')
// …
mix.options({
  sourcemaps: mix.inProduction() ? false : 'source-map'
})
```

thecrypticace:
Closes #2600

@JeffreyWay Should we mark that config option as internal and/or remove it from public API?

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -34,7 +34,7 @@ interface MixConfig {
         onSuccess?: boolean;
         onFailure?: boolean;
     };
-    sourceMaps?: boolean;
+    sourcemaps?: false | string;
     resourceRoot?: string;
     imgLoaderOptions?: {
         enabled?: boolean;


JeffreyWay:
Yeah, we could. Though I can't think of any place that we recommend the approach he used in that issue. All the docs say mix.sourceMaps();

roelofr:
Maybe I could make a PR on the correct (`mix.sourceMaps`) use, since it's only mentioned briefly in `examples.md` at this point.

JeffreyWay:
@roelofr That'd be great. 👍

PR #2609, opened 2020-11-08
Remove need for cross-env

thecrypticace:
Hoping this works 🤞 

Using CI to do some Windows testing.

Current implementation relies on #2608 but can easily be based on `master`


JeffreyWay:
Is there a benefit to removing cross-env? Or is it just a "one less dependency" thing?

thecrypticace:
I’m still thinking about whether or not this is a good idea. The main improvement here is simplification for windows users which is the whole reason we need to use cross-env. After thinking about this though it may still prove useful.

thecrypticace:
After thinking about this more I'm gonna table it for now. I may re-open it in the future if it makes sense to do so.

PR #2622, opened 2020-11-21
Quote paths when building CLI script command.

### LINKED ISSUE #2621: Laravel mix cannot find config file
-   Laravel Mix Version: 6.0.0-beta.14 (`npm list --depth=0`)
-   Node Version (`node -v`): v12.19.1
-   NPM Version (`npm -v`): 6.14.8
-   OS: Pop!_OS 20.04 LTS

### Description:
I am testing out the beta release of laravel mix, but every time I try to compile my assets it throws an error saying it cannot find my config file. `[webpack-cli] The specified config file doesn't exist in /media/storage/Work`
It seems like something isn't passing the current path correctly since the path mentioned above isn't actually the full path to where I am compiling my code from. This is my actual CWD: `/media/storage/Work Files/Code/laravel-mix`

When I tested this, everything worked as expected if I ran it from the following path: `/tmp/laravel-mix` which doesn't contain any spaces in its full path.

### Steps To Reproduce:

1. Create a folder with a space in the name.
2. Try to compile code with laravel mix `npx mix`

### webpack.mix.js file
```js
let mix = require('laravel-mix');

mix.postCss('./resources/css/main.css', 'css')

mix.webpackConfig({
    externals: {
        'jquery': 'jQuery',
    },
})

mix.options({
    processCssUrls: false,
    postCss: [
        require('postcss-import'),
        require('tailwindcss'),
        require('postcss-nested'),
    ]
}).setResourceRoot('..')
    .setPublicPath('./assets');

if (mix.inProduction()) {
    mix.sourceMaps();
}
`

ericp-mrel:
Fixes #2621 

Make sure paths are quoted when building the CLI script command in case any paths contain spaces.

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -87,10 +87,12 @@ function commandScript(cmd, opts) {
  */
 function commandArgs(args) {
     return args.map(arg => {
-        const keyValue = arg.split('=');
+        // Split string at first = only
+        const pattern = /^([^=]+)=(.*)$/;
+        const keyValue = arg.includes('=') ? pattern.exec(arg).slice(1) : [];
 
-        if (typeof keyValue[1] !== 'undefined') {
-            arg = `${keyValue[0]}="${keyValue[1]}"`;
+        if (keyValue.length === 2) {
+            return `${keyValue[0]}="${keyValue[1]}"`;
         }
 
         return arg;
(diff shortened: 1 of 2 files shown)


thecrypticace:
Ah yeah, this is definitely a problem. Thanks for this!

thecrypticace (approved):

ericp-mrel:
@thecrypticace I've added some additional fixes with regards to passing extra arguments to webpack so if someone passes additional env arguments to webpack those values will be quoted now as well since those were also being passed unquoted.

This should hopefully take care of all of the unquoted CLI command I can think of at this moment.

thecrypticace (approved):

JeffreyWay:
Thank you, @ericp-mrel!!

PR #2179, opened 2019-07-29
Remove notification timeout, fixes #2161

AdrienPoupa:
Remove timeout notification, this breaks the notifications on KDE

https://github.com/JeffreyWay/laravel-mix/issues/2161

PR #2626, opened 2020-11-23
Remove default 10 second timeout from notifications

kinglozzer:
Regression from #2179 - terminal hangs for 10 seconds after a build has finished before allowing the user to type again.

Quoting the [node-notifier docs](https://github.com/mikaelbr/node-notifier#all-notification-options-with-their-defaults):

> As of Version 6.0 there is a default timeout set of 10 to ensure that the application closes properly. In order to remove the timeout and have an instantly closing notification (does not support actions), set timeout to false. If you are using action it is recommended to set timeout to a high value to ensure the user has time to respond.

“Instantly closing” appears to be misleading here, as the notification stays for ~4 seconds on MacOS 11.0.1. If that behaviour isn’t consistent cross-platform, I think it’d be best to revert #2179 and instead track the milliseconds vs seconds issue in `node-notifier` rather than try to fix it in this project: https://github.com/mikaelbr/node-notifier/issues/287

@AdrienPoupa @vokamut @rodrigopedra I notice you were all involved in testing the PR that removed `timeout: 2`, could I ask you to please test this change if you have a spare moment? Thanks!


JeffreyWay:
Thanks @kinglozzer. I'm going to merge this. Still not sure if the instantly closing delay is consistent cross platform, but someone can create a PR if so.

PR #2632, opened 2020-11-29
Update webpack-dev-server dependency

nyxsoftwaresystems:
This should fix numerous dependency errors that others are experiencing.


thecrypticace (approved):
Thanks for this!

PR #2642, opened 2020-12-01
[6.x] Install dependencies with yarn when using yarn

### LINKED ISSUE #2614: Support for yarn & don't assume --save-dev
-   Laravel Mix Version: 5.0.7
-   Node Version: 10.19.0
-   NPM Version: 6.13.4
-   Yarn Version: 1.22.4
-   OS: Win10 1909

### Description 1:
When running the webpack bundler using mix as the config wrapper and you have missing dependencies, mix automatically tries to install dependencies but assumes you are using npm as the package manager. We use yarn & need to cancel the build in order to manually install the missing dependencies.

```
> cross-env NODE_ENV=development node_modules/webpack/bin/webpack.js --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js "--watch"

Additional dependencies must be installed. This will only take a moment.

Running: npm install sass --save-dev --production=false
```

### Description 2:
Mix assumes you want to install the missing dependencies using flags `--save-dev --production=false`. This assumption should not be made if you are automating steps on behalf of the user.

Since Laravel is not a nodejs application, none of the node packages are used in production, making the distinction between prod & dev arbitrary. We only install packages used for testing as dev dependencies, the rest as prod dependencies.

### Steps To Reproduce:
1. Run `yarn install` to generate your `yarn.lock` file.
2. Make sure one of the dependencies is missing (in our example the `sass` package).
3. Run the bundler `cross-env NODE_ENV=development node_modules/webpack/bin/webpack.js --progress --hide-modules --config=node_modules/laravel-mix/setup/webpack.config.js "--watch"`

thecrypticace:
I actually had this code laying around in a branch and forgot to PR it. ¯\\_(ツ)_/¯  

I can backport this to v5 too if you like.

Fixes #2614

### DIFF
--- a/src/Dependencies.js
+++ b/src/Dependencies.js
@@ -81,6 +81,9 @@ class Dependencies {
         switch (PackageManager.detect()) {
             case 'npm':
                 return `npm install ${dependencies} --save-dev --legacy-peer-deps`;
+
+            case 'yarn':
+                return `yarn add ${dependencies} --dev`;
         }
     }
 
--- a/src/Dependencies.js
+++ b/src/Dependencies.js
@@ -1,6 +1,7 @@
 let childProcess = require('child_process');
 let Log = require('./Log');
 let argv = require('yargs').argv;
+let PackageManager = require('./PackageManager');
 
 /**
  * @typedef {object} DependencyObject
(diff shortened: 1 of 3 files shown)


thecrypticace:
I'm pretty sure the windows failure is not a real failure but some problem with the runner. :/

smujaddid:
This PR forced us to automatically install dependecies with yarn when my package-lock.json is based on npm.
I'm using both yarn and npm but for different projects.

Is there a way to force using npm instead of yarn when both are installed?

thecrypticace:
You're using both yarn and npm? How does that work? Can you give more details on your setup?

PR #2655, opened 2020-12-05
[React] add support to new JSX transform

ecreeth:
- With the new transform, you can use JSX without importing React.
- Depending on your setup, its compiled output may slightly improve the bundle size.

Additional information can be found here:

- [Reactjs.org](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html)
- [Babel](https://babeljs.io/blog/2020/03/16/7.9.0)

### DIFF
--- a/src/components/React.js
+++ b/src/components/React.js
@@ -23,7 +23,7 @@ class React {
      */
     babelConfig() {
         return {
-            presets: ['@babel/preset-react']
+            presets: [['@babel/preset-react', { runtime: 'automatic' }]]
         };
     }
 }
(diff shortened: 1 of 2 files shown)


src/components/React.js
```diff
@@ -23,7 +23,7 @@ class React {
      */
     babelConfig() {
         return {
-            presets: ['@babel/preset-react']
+            presets: [['@babel/preset-react', { runtime: 'automatic' }]]
```
ecreeth, line 26: **Note**: Starting from Babel 8, "automatic" will be the default runtime.

thecrypticace (approved):

PR #2660, opened 2020-12-07
Apply webpack-dev-server 4.0 changes

### LINKED ISSUE #2653: [6.*] npx mix watch --hot gives Unknown argument
-   Laravel Mix Version: 6.0.0-beta.15 (`npm list --depth=0`)
-   Node Version (`node -v`): 14.15.1
-   NPM Version (`npm -v`): 6.14.8
-   OS: windows

### Description:
I get this error upon running the command
> mix watch --hot
[webpack-cli] Unknown argument: --inline
[webpack-cli] Unknown argument: --disable-host-check

### Steps To Reproduce:
npm install laravel-mix@6.0.0-beta.15 -D
Then npx mix watch --hot

soilSpoon:
This PR resolve #2653

## Reference
[webpack/webpack-dev-server/releases/tag/v4.0.0-beta.0](https://github.com/webpack/webpack-dev-server/releases/tag/v4.0.0-beta.0)

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -91,7 +91,7 @@ function commandScript(cmd, opts) {
 
         return 'npx webpack --watch';
     } else if (cmd === 'watch' && opts.hot) {
-        return 'npx webpack serve --hot --inline --disable-host-check';
+        return 'npx webpack serve --hot';
     }
 }
 
@@ -117,7 +117,6 @@ function quoteArgs(args) {
 function isTesting() {
     return process.env.TESTING;
 }
-
 
 function isTTY() {
     if (isTesting() && process.env.IS_TTY !== undefined) {
--- a/src/builder/webpack-default.js
+++ b/src/builder/webpack-default.js
@@ -7,6 +7,8 @@ module.exports = function() {
 
         mode: Mix.inProduction() ? 'production' : 'development',
 
+        infrastructureLogging: Mix.isWatching() ? {} : { level: 'none' },
+
         entry: {},
 
         output: {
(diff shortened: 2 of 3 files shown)


thecrypticace:
From what I can tell this looks good to me. You can ignore the windows failure right now. Something is causing weird issues there that I have to look into.

I plan to add an integration test for the webpack-cli so this could've been caught earlier.

bin/cli.js
```diff
@@ -91,7 +91,7 @@ function commandScript(cmd, opts) {
 
         return 'npx webpack --watch';
     } else if (cmd === 'watch' && opts.hot) {
-        return 'npx webpack serve --hot --inline --disable-host-check';
+        return 'npx webpack serve --hot';
```
soilSpoon, line 94: The `hot` option is left over because it checks if it is included in `process.argv`.
Let me know if there is a better way

thecrypticace: We should probably ask webpack itself if it's running via HMR. I'll take a look today to see how to do that.

oza75: It will also be good if we can enable the webpack-dev-server https. I tried this `npx mix --hot --https` but got an` unknow option` error. it would be nice if we weren't limited to the options defined by laravel-mix

thecrypticace: We're not! You can pass arbitrary options after `--` like so `npx mix --hot -- --https`

oza75: Thanks @thecrypticace. it worked but when I reload my page I get net::ERR_CERT_AUTHORITY_INVALID http error. Maybe I'm doing something wrong

thecrypticace: That's because webpack's https support uses a self-signed certificate by default. If you want to customize that you'll have to customize the webpack config.

JeffreyWay:
Thanks for this @soilSpoon. I'll go ahead and merge this so it can go out as part of the next beta I'm about to release.

PR #2661, opened 2020-12-07
Add "Fast Refresh" to React Component

soilSpoon:
Fast Refresh is a feature that lets you edit React components in a running application without losing their state. It is similar to an old feature known as "hot reloading", but Fast Refresh is more reliable and officially supported by React.

## Reference
[pmmmwh/react-refresh-webpack-plugin](https://github.com/pmmmwh/react-refresh-webpack-plugin)
[react-refresh](https://github.com/facebook/react/tree/master/packages/react-refresh)


src/components/React.js
```diff
@@ -3,7 +3,16 @@ class React {
...
      */
     dependencies() {
-        return ['@babel/preset-react'];
+        const dependencies = ['@babel/preset-react'];
+
+        if (Mix.isHot()) {
+            dependencies.concat([
+                '@pmmmwh/react-refresh-webpack-plugin',
+                'react-refresh'
+            ]);
```
thecrypticace, line 12: I'm inclined to say we should always include these.

thecrypticace:
I'l leave this for @JeffreyWay to decide but if he's cool with adding it we'll want to add tests for it.

src/components/React.js
soilSpoon: Why? I think this is only necessary when using HMR.

thecrypticace: Because once they're added to the package.json file they'll be in there whether you're using it or not. It's fine either way honestly I just think it'd be less hassle.

hivokas: image

soilSpoon:
I should do this if `React` pacakge version is 16.10 or higher

## Reference
https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/scripts/start.js

soilSpoon:
How should I check the react version? The solution I think is to use [semver](https://www.npmjs.com/package/semver).
Or is it the right way not to add this to this package?

thecrypticace:
You can grab the react version, when installed, by doing this: `require("react").version`

We'd need to do an actual version comparison. A quick google search says we could use the `compare-versions` package for that.

src/components/React.js
```diff
@@ -2,10 +2,22 @@ class React {
     /**
      * Required dependencies for the component.
      */
+
+    supportFastRefresh() {
+        try {
+            const semver = require('semver');
+            const react = require('react');
+
+            return Mix.isHot() && semver.satisfies(react.version, '^16.9');
```
thecrypticace, line 11: I think you'd want to check for `^16.9 || ^17.0`, yeah?

JeffreyWay:
If this functionality is very common in the React community, I'm happy to merge this. 

@soilSpoon Any idea what kind of tests we could put behind this?

thecrypticace:
It's becoming very common. It's enabled by default in in Create React App. And Vite has a plugin specifically to provide support for it. It's more powerful version of HMR for React components. It is the successor to React's old hot reloading package `react-hot-loader`.

soilSpoon:
@JeffreyWay Optionally, Fast Refresh works well, so you can test whether the screen does not flicker and the modifications are reflected.
But i don't know what to do with that test.
And it's good to test if it doesn't work in React `<16.9`, but I don't know what to do.

thecrypticace:
A test that the plugins and presets are added for React >= 16.9 and not for < 16.9 would be a good place to start. Integration tests for HMR would be a bit more complicated and not something to tackle right now.

soilSpoon:
What else do I need to do?

JeffreyWay:
Okay I'll merge this and add the tests myself. Thanks @soilSpoon.

PR #2663, opened 2020-12-09
Refactor test to check the PostCSS version

soilSpoon:
PostCSS 8.2 has been released, so using String.prototype.startsWith() throws an error.
This PR resolves the problem.

## Reference
https://github.com/postcss/postcss/releases/tag/8.2.0

### DIFF
--- a/test/unit/Dependencies.js
+++ b/test/unit/Dependencies.js
@@ -2,6 +2,7 @@ import '../../src/helpers';
 import test from 'ava';
 import childProcess from 'child_process';
 import sinon from 'sinon';
+import semver from 'semver';
 import Dependencies from '../../src/Dependencies';
 import PackageManager from '../../src/PackageManager';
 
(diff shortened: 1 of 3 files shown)


test/unit/Dependencies.js
```diff
@@ -72,7 +73,7 @@ test('it can utilize custom checks for a dependency', t => {
             check: postcss => {
                 called = true;
 
-                t.true(postcss().version.startsWith('8.1'));
+                t.true(semver.satisfies(postcss().version, '^8.1'));
```
thecrypticace, line 76: You know… if I'd've known we already had the semver package installed I would've used it. Definitely a much better option.

👍

thecrypticace:
Can you also update `src/components/CssWebpackConfig.js` while you're at it?

test/unit/Dependencies.js
soilSpoon: I know. So adding semver looks very good.

thecrypticace:
Thanks for this!

PR #2672, opened 2020-12-11
[6.x] Remove SnoreToast message from notification in Windows 10

### LINKED ISSUE #2671: SnoreToast is added to the start menu when notifying on Windows 10
-   Laravel Mix Version: 5.0.9
-   Node Version: v15.4.0
-   NPM Version: 7.0.15
-   OS: win10-x64

### Description:
Untitled

Every time mix compiles [snoretoast](https://github.com/KDE/snoretoast) is always added to the start menu as a shortcut and creates a registry key which is very annoying. I think this is affecting both 5.x and 6.x

The same issue was reported from vue-cli [here](https://github.com/vuejs/vue-cli/issues/2720) and here is the [fix](https://github.com/vuejs/vue-cli/pull/6000).

However, the difference is that vue-cli uses [node-notifier](https://github.com/mikaelbr/node-notifier) but mix uses [webpack-notifier](https://github.com/Turbo87/webpack-notifier).

The issue was also reported in [node-notifier repo issue](https://github.com/mikaelbr/node-notifier/issues/154)

I also fired an issue at [webpack-notifier](https://github.com/Turbo87/webpack-notifier/issues/64) see what we can do.
### Steps To Reproduce:
Windows 10 run npm run dev

thecrypticace:
Fixes #2671

### DIFF
--- a/src/components/Notifications.js
+++ b/src/components/Notifications.js
@@ -9,15 +9,12 @@ class Notifications extends AutomaticComponent {
             let WebpackNotifierPlugin = require('webpack-notifier');
 
             return new WebpackNotifierPlugin({
+                appID: 'Laravel Mix',
+
                 title: 'Laravel Mix',
                 alwaysNotify: Config.notifications.onSuccess,
-                hint:
-                    process.platform === 'linux'
-                        ? 'int:transient:1'
-                        : undefined,
-                contentImage: Mix.paths.root(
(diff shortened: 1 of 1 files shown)


Nothing-Works:
Can this be in both v5.x and v6.x, please?

thecrypticace:
Opened #2673 for Mix v5

PR #2693, opened 2020-12-22
Only check filename part.

danny50610:
In Windows:  
```
process.env.npm_execpath = "C:\Program Files (x86)\Yarn\bin\yarn.js"
or
"C:\Program Files\nodejs\node_modules\npm\bin\npm-cli.js"
```

I think just checking the filename is enough. Otherwise the slash of the path is very annoying.

### DIFF
--- a/src/PackageManager.js
+++ b/src/PackageManager.js
@@ -1,12 +1,21 @@
+const File = require('./File');
+const Paths = require('./Paths');
+
 class PackageManager {
     static detect() {
         const execPath = process.env.npm_execpath || '';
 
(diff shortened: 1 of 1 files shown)


thecrypticace:
This looks good to me. I also pushed a commit to check for the presence of a yarn.lock file because I discovered that if you supply something like `"dev": "npx mix"` rather than `"dev": "mix"` in your package.json scripts and you run `yarn dev` the check doesn't work as the exec path will be the npx CLI.

thecrypticace (approved):

PR #2694, opened 2020-12-23
Bump prettier and pretty-quick

soilSpoon:
Bump `prettier` and `pretty-quick` and apply to files that do not have `prettier` applied.


JeffreyWay:
Thanks!

PR #2700, opened 2020-12-23
Add node browser polyfills for process and Buffer

### LINKED ISSUE #2662: [6.x] Legacy Node Polyfills
We should either add a config option / component that re-enables these or a note to the docs.

What needs to happen is rather simple:

1. Install `node-libs-browser`
2. Update the webpack config
```js
mix.webpackConfig({
    resolve: {
        fallback: {
            // This handles all of them…
            // ...Object.fromEntries(Object.entries(require("node-libs-browser")).filter(e => e[1] !== null)),

            // … but these are likely the most used and as such should be sufficient
            process: require("node-libs-browser").process,
            buffer: require("node-libs-browser").buffer,
        },
    },
})
```

If we were to add a shortcut for this in Mix I'd probably do something like this:
```js
mix.options({ legacyNodePolyfills: true })
```

What do you think @JeffreyWay? Would documentation suffice or should we add a shortcut for this?

### LINKED ISSUE #2685: process.ENV blank in v6
Previous versions when we had a .env file we could do things such as...

MIX_API_KEY=123456

Then access it via JavaScript using 

process.env.MIX_API_KEY

however this is now coming up as undefined, I'm assuming this is because webpack removed it by default? https://webpack.js.org/migrate/5/ says to use DefinePlugin or EnvironmentPlugin

Could we bundle that in for compatibility?

thecrypticace:
Fixes #2662
Fixes #2685

### DIFF
--- a/src/components/LegacyNodePolyfills.js
+++ b/src/components/LegacyNodePolyfills.js
@@ -1,0 +1,34 @@
+const AutomaticComponent = require('./AutomaticComponent');
+const webpack = require('webpack');
+
+class LegacyNodePolyfills extends AutomaticComponent {
+    webpackPlugins() {
+        if (!Config.legacyNodePolyfills) {
+            return [];
+        }
+
+        return [
+            new webpack.ProvidePlugin({
+                Buffer: 'buffer',
+                process: 'process/browser'
+            })
+        ];
+    }
+
+    webpackConfig() {
+        if (!Config.legacyNodePolyfills) {
+            return {};
+        }
+
+        return {
+            resolve: {
+                fallback: {
+                    Buffer: 'buffer',
+                    process: 'process/browser'
+                }
+            }
+        };
+    }
+}
+
+module.exports = LegacyNodePolyfills;
--- a/package.json
+++ b/package.json
@@ -82,6 +82,7 @@
         "lodash": "^4.17.20",
         "md5": "^2.3.0",
         "mini-css-extract-plugin": "^1.1.0",
+        "node-libs-browser": "^2.2.1",
         "postcss-load-config": "^3.0.0",
         "postcss-loader": "^4.0.4",
         "semver": "^5.7.1",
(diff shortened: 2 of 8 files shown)


src/components/LegacyNodePolyfills.js
```diff
@@ -0,0 +1,30 @@
...
+        return [
+            new webpack.ProvidePlugin({
+                Buffer: 'buffer',
+                process: 'process/browser'
+            })
+        ];
+    }
+
+    webpackConfig() {
+        return {
```
JeffreyWay, line 19: This will be added regardless of if `Config.legacyNodePolyfills` is set. Is that okay? Any issue there?

thecrypticace: Heh oops. Fix incoming

JeffreyWay:
Perfect. Thanks!

soilSpoon:
`node-libs-browser` is deprecated. so i think using this library could hurt UX.
Therefore i think it should change from webpack/node-libs-browser to feross/buffer and shtylman/node-process.

What do you think about the above?

thecrypticace:
Yeah we should probably do that. `node-libs-browser` is, in reality, a package that serves only to include other packages + provide a unified way to get what import is useful for the browser. I missed the note that it is deprecated.

Kaos-Industries:
I can only get this solution working if I do the following:

```
mix
	.js("resources/js/app.js", "public/js")
	.js("resources/js/payments.js", "public/js")      // Had to add this 
	.sass("resources/sass/app.scss", "public/css")
	.postCss("resources/css/app.css", "public/css")
	.sourceMaps()
	.options({
		legacyNodePolyfills: true,      // And this
	});
```

...where `payments.js` is the file I'm trying to access my Laravel environment variables from. This isn't ideal because, at least without npm's `--production` flag, the `payments.js` file gets "compiled" to a size that's about ten times bigger than the pre-compiled file. Is there a way to get this fix working without that extra .js compilation line in my `webpack.mix.js`?

PR #2708, opened 2020-12-25
Fix windows failures

thecrypticace:
Testing to see if I can figure out why the node 14 win failure is consistently (but still sometimes intermittently) happening


thecrypticace:
I'mma leave this open for now while I ponder what might be wrong. I think it's Ava related. I think one of the test processes is getting killed towards the end.

PR #2712, opened 2020-12-26
Add tests for async style-containing vue components when extracting styles

thecrypticace:
This adds tests for extraction of styles from async vue components. We could not import these components previously due to an issue in webpack. This appears to have been resolved in v5.25.1.

### DIFF
--- a/src/components/Preprocessor.js
+++ b/src/components/Preprocessor.js
@@ -103,7 +103,7 @@ class Preprocessor {
     /**
      * Prepare the preprocessor plugin options.
      *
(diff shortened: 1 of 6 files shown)


JeffreyWay:
@thecrypticace Can we split these up into their own PRs, so I can look over and merge the first fix you've implemented here?

thecrypticace:
Yeah, will do. I need to rebase this. All of this is really just the one workaround for async Vue component styles + a fix for styles from JSX.

PR #2713, opened 2020-12-27
Update MergeSmart.js according to eslint errors.

halilcakar:
eslint(newline-before-return) errores fixed.


halilcakar:
Well i didn't get why it's failed. Since all I've done just added some spaces 😁

thecrypticace:
Heh, it's a flaky job. Not your fault! It's always the win10 Node 14 build that fails (… but not always).

halilcakar:
Thanks for answer @thecrypticace. 

What do you think we can do about it. This was just a eslint error fix nothing more. Doesn't change how code works or smt.

PR #2714, opened 2020-12-28
[DNM] Log segfaults to stderr

thecrypticace:
So the intermittent playwright error is now gone but Node 14 on Windows 10 has seen problems with access violations on Win 10 (see https://github.com/nodejs/help/issues/2660)

They supposedly disappeared but seeing that we're having the same problem that appears to not (or no longer) be the case.

I'm attempting to capture a stack trace here when it crashes to see if we need to file a bug with Node. We'll see if that's actually possible. If it's not I'll close this PR and work on this another day as we've got some other Mix bugs to take care of.

### DIFF
--- a/package.json
+++ b/package.json
@@ -122,6 +122,7 @@
         "sass": "^1.27.0",
         "sass-loader": "^10.0.3",
         "sass-resources-loader": "^2.1.1",
+        "segfault-handler": "^1.3.0",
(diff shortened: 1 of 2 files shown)


thecrypticace:
Sigh access violations don't get caught here. 😭 
Will tackle another day.

PR #2724, opened 2020-12-30
Table output enhancements

### LINKED ISSUE #2721: Compiled assets table has some problems
-   Laravel Mix Version: 6.0.5
-   Node Version: v14.15.3
-   NPM Version: 6.14.10
-   OS: Ubuntu Server 20.04 x64

### Description:

1) New table layout is less graphic than the previous "list" layout. Also it's taking too much vertical space and too little horizontal space:

gh_shot_1

2) Files are sorted by size, not by name. It is not very convenient if there are a lot of files in the project. For example, this happens if I simply copy some directory from `node_modules` to `public`:

gh_3

Compare with the old layout:

gh_4

There is much more information taking up much less space (one single screen in my case, vs. **24** screens (!) in table layout).

The same project built with Mix 6:

gh_5

Also `app.css.gz` and `app.css.map` are missing (though they do exist in my `public` directory). It is now unclear whether they are being created at all.

## TL;DR

My suggestions are:

- Make the table fill all available width
- Don't limit the file name length (it's important to know the *ending* of the file path, not the *beginning*)
- Sort files by name, not by size (or maybe *sort* by size, but *group* by name)
- Fix the problem when some files are not displayed, although they should (e. g. `.map` files or files handled by webpack plugins)

thecrypticace:
Mostly fixes #2721

1. The table will now fill the terminal width
2. Files are sorted by name
3. Files are truncated at the beginning of their path — not the end
4. The file column is now right aligned
5. The table is now drawn compactly so there's no separator between file rows. I've noticed on a few of my projects too with more than a handful of files that taking up all that space w/ the table can be fairly overwhelming to look at.

I still need to look into why some files are not displayed. That doesn't necessarily preclude this being merged though. That can be a separate PR.


JeffreyWay:
@thecrypticace Looks good. Any thoughts on the Ubuntu failures?

thecrypticace:
I'm gonna take a look. I think it's either related to the snapshot or for some reason it's not honoring setting process.stdout.columns on linux. I need to setup a docker container to run the tests on linux and see if I see any behavioral differences.

thecrypticace:
Figured it out. Was related to snapshots combined with how ANSI escape codes are handled differently across platforms.

soilSpoon:
How do you think about aligning the table left and giving the width dynamic according to the content?
I think it looks better.

thecrypticace:
1. Table data was truncated on the wrong side of the table. You only saw the beginning of the path. Left-aligning data truncated at the start scans weird.
2. It matches the webpack output

We absolutely should handle large monitors though. It is crazy wide. Maybe a max of 120-ish cols and make it configurable?

soilSpoon:
``` javascript
const _ = require('lodash');
const chalk = require('chalk');
const Table = require('cli-table3');
const readline = require('readline');
const stripAnsi = require('strip-ansi');
const { formatSize } = require('webpack/lib/SizeFormatHelpers');
const { version } = require('../../package.json');

/**
 * @typedef {object} BuildOutputOptions
 * @property {boolean} clearConsole
 * @property {boolean} showRelated
 **/

/**
 * @typedef {object} StatsAsset
 * @property {string} name
 * @property {number} size
 * @property {StatsAsset[]|{}} related
 **/

/**
 * @typedef {object} StatsData
 * @property {StatsAsset[]} assets
 **/

class BuildOutputPlugin {
    /**
     *
     * @param {BuildOutputOptions} options
     */
    constructor(options) {
        this.options = options;
        this.patched = false;
    }

    /**
     * Apply the plugin.
     *
     * @param {import("webpack").Compiler} compiler
     */
    apply(compiler) {
        // TODO: Refactor setup to allow removing this check
        if (process.env.NODE_ENV === 'test') {
            return;
        }

        compiler.hooks.done.tap('BuildOutputPlugin', stats => {
            if (stats.hasErrors()) {
                return false;
            }

            if (this.options.clearConsole) {
                this.clearConsole();
            }

            let data = stats.toJson({
                assets: true,
                builtAt: true,
                hash: true,
                performance: true,
                relatedAssets: this.options.showRelated
            });

            this.heading(`Laravel Mix v${version}`);

            console.log(chalk.green.bold(`✔ Compiled Successfully in ${data.time}ms`));

            if (data.assets.length) {
                console.log(this.statsTable(data));
            }
        });
    }

    /**
     * Print a block section heading.
     *
     * @param {string} text
     */
    heading(text) {
        console.log();

        console.log(chalk.bgBlue.white.bold(this.section(text)));

        console.log();
    }

    /**
     * Create a block section.
     *
     * @param {string} text
     */
    section(text) {
        const padLength = 3;
        const padding = ' '.repeat(padLength);

        text = `${padding}${text}${padding}`;

        const line = ' '.repeat(text.length);

        return `${line}\n${text}\n${line}`;
    }

    /**
     * Generate the stats table.
     *
     * @param {StatsData} data
     * @returns {string}
     */
    statsTable(data) {
        const assets = this.sortAssets(data);

        const table = new Table({
            head: [chalk.bold('File'), chalk.bold('Size')],
            style: {
                head: [],
                compact: true
            }
        });

        for (const asset of assets) {
            table.push([chalk.green(asset.name), formatSize(asset.size)]);
        }

        this.monkeyPatchTruncate();

        return table.toString();
    }

    /**
     *
     * @param {StatsData} data
     */
    sortAssets(data) {
        let assets = data.assets;

        assets = _.flatMap(assets, asset => [
            asset,
            ...(Array.isArray(asset.related) ? asset.related : [])
        ]);

        assets = _.orderBy(assets, ['name', 'size'], ['asc', 'asc']);

        return assets;
    }

    /**
     * Clear the entire screen.
     */
    clearConsole() {
        const blank = '\n'.repeat(process.stdout.rows);
        console.log(blank);

        readline.cursorTo(process.stdout, 0, 0);
        readline.clearScreenDown(process.stdout);
    }

    // Yeah, I know.
    monkeyPatchTruncate() {
        if (this.patched) {
            return;
        }

        this.patched = true;

        // @ts-ignore
        const utils = require('cli-table3/src/utils');
        const oldTruncate = utils.truncate;

        // cli-table3 can only do truncation at the end
        // We want the asset name to be truncated at the beginning if it's too long
        // FIXME: We really should set truncation location via a paramter or something
        // (or PR support for alignment-based truncation)

        /**
         *
         * @param {string} str
         * @param {number} desiredLength
         * @param {string} truncateChar
         */
        utils.truncate = (str, desiredLength, truncateChar) => {
            if (stripAnsi(str).length > desiredLength) {
                str = `…${str.substr(-desiredLength + 2)}`;
            }

            return oldTruncate(str, desiredLength, truncateChar);
        };
    }
}

module.exports = BuildOutputPlugin;

```
How do you think of modifying the `src/webpackPlugins/BuildOutputPlugin.js` file as above to make it dynamic width with Left alignment?

### Before
image

### After
image

isotrope:
Please. I'm begging you, let use use --hide-module or the equivalent like before.
As with #2764 I'm getting two screens and more of output, which makes it impossible to even detect if I'm looking at an old results table, a new one.
This might seem like a dumb detail but it's probably going to force me to replace laravel-mix with another build methodology. :(

End time of compilation would be amazing (to know if it's done or not, or if we're looking at an older build).

lk77:
Hello,

same issue here, 

i tried :

```
stats: {
            preset: 'minimal', 
}
```

in .webpackConfig() but it does not work.

PR #2727, opened 2020-12-30
Switch to vue-style-loader when using mix.vue()

### LINKED ISSUE #2725: Add vue-style-loader for shadowMode
-   Laravel Mix Version: 6.0.5
-   Node Version: v12.14.1
-   NPM Version: 6.13.4
-   OS: Windows10

### Description:
Based on [this](https://github.com/JeffreyWay/laravel-mix/issues/2643)

With new version we can pass `.vue({ options })`, 1 of the options of `vue-loader` is `shadowMode`, but shadowMode doesn't work without `vue-style-loader`.

It is possible to add `option` to change default `style-loader` to `vue-style-loader`, I was thinking something like this:

```javascript
// config.js

...
shadowMode: false,
...
```

```javascript
// components/CssWebpackConfig.js

static afterLoaders({ method = 'auto' } = {}) {
    const loaders = [];

    if (Config.shadowMode) {
        loaders.push({
            loader: 'vue-style-loader',
            options: {
                shadowMode: true
            }
        });

        return loaders;
    }

    if (method === 'auto') {
        // TODO: Fix
        if (Mix.extractingStyles !== false) {
            method = 'extract';
        } else {
            method = 'inline';
        }
    }

    if (method === 'inline') {
        loaders.push({ loader: 'style-loader' });
    } else if (method === 'extract') {
        loaders.push({
            loader: MiniCssExtractPlugin.loader,
            options: {
                esModule: true
            }
        });
    } else {
        throw new Error(
            `Unknown css loader method '${method}'. Expected auto, inline, or extract.`
        );
    }

    return loaders;
}
```
```javascript
//mixFile
mix.options({
	shadowMode: true
});

mix.js('src/index.js', 'dist/app.js').vue({
	version: 2,
	options: {
		shadowMode: true
	}
});
```

Ofcourse it will be more correct to use the `.vue({ options })`, but I couldn't find how I can access vue options in `components/CssWebpackConfig.js`

thecrypticace:
Fixes #2725

### DIFF
--- a/src/components/CssWebpackConfig.js
+++ b/src/components/CssWebpackConfig.js
@@ -171,8 +171,9 @@ class CssWebpackConfig extends AutomaticComponent {
      *
      * @param {object} [options]
      * @param {"auto" | "inline" | "extract"} options.method The method to use when handling CSS.
+     * @param {"default" | "per-file"} options.location Where these loaders are applied. The `default` set or on a per-file basis (used by preprocessors).
      */
-    static afterLoaders({ method = 'auto' } = {}) {
+    static afterLoaders({ method = 'auto', location = 'default' } = {}) {
         const loaders = [];
 
         if (method === 'auto') {
@@ -185,7 +186,11 @@ class CssWebpackConfig extends AutomaticComponent {
         }
 
         if (method === 'inline') {
-            loaders.push({ loader: 'style-loader' });
+            if (Mix.components.get('vue') && location === 'default') {
+                loaders.push({ loader: 'vue-style-loader' });
+            } else {
+                loaders.push({ loader: 'style-loader' });
+            }
         } else if (method === 'extract') {
             loaders.push({
                 loader: MiniCssExtractPlugin.loader,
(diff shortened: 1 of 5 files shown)


Froxz:
Hi @thecrypticace,

I think you should keep `style-loader` as default, becuase `vue-style-loader`, doesn't support `css-loader` 4+ see [here](https://github.com/vuejs/vue-style-loader/pull/47)

thecrypticace:
Yep, saw that this morning. ￼Pretty sure that's why the test fails. I'll keep this around to have it ready when vue-style-loader updates￼.

Froxz:
> Yep, saw that this morning. ￼Pretty sure that's why the test fails. I'll keep this around to have it ready when vue-style-loader updates￼.

emmm I don't think they will update, that PR is hanging from Aug

Froxz:
Hi @thecrypticace 

vue-style-loader (v4.1.3) was updated to support css-loader 4+, can we now try to implement the above?

Froxz:
Hi @thecrypticace the build test failed again, but I don't think is related to `vue-style-loader`

PR #2730, opened 2020-12-30
Install all dependencies at once

### LINKED ISSUE #2728: Install all dependencies at once
One annoying quirk (my fault) is that Mix doesn't install all required dependencies in one go. Instead, when it runs into a missing dependency, it stops and immediately runs the install command. This means you might have to run mix two or more times to pull in all required dependencies for your current webpack.mix.js file. 

Screen Shot 2020-12-30 at 10 44 28 AM

Instead, I'd prefer that it cycle through all relevant components and build up a single install command for everything.

thecrypticace:
Fixes #2728


setup/webpack.config.js
```diff
@@ -2,6 +2,8 @@ module.exports = async () => {
     require('../src/index');
     require(Mix.paths.mix());
 
+    await Mix.dispatch('internal:gather-dependencies');
+    await Mix.dispatch('internal:install-dependencies');
```
thecrypticace, line 6: I named these this way because we might want to refactor this implementation later and I would not want extensions to rely on these events as they're an implementation detail.

thecrypticace:
Converting this to a draft. I have some in progress ideas for cleanups / refactoring and I think I can land a very, very small portion of that which would make the approach for this less iffy.

PR #2749, opened 2021-01-06
Table output enhancements take 2

thecrypticace:
Now with the fix that accounts for ANSI escape codes.


florianbouvot:
@thecrypticace new table are better, thanks.

thecrypticace:
Awesome, good to hear! 👍

PR #2797, opened 2021-01-24
Optimize BabelConfig at the top level

soilSpoon:
There was no way to specify the babel config file, so an option was added to the config.
Also, the babel config merge order in src/babelConfig.js has been modified so that the user can modify the babel config set in the component.

## Aside
It would be nice to automatically detect the [general names of babel config](https://babeljs.io/docs/en/config-files#configuration-file-types) files in the future.


thecrypticace:
In general this looks good. There's some breaking changes here but I _think_ they're probably not that big of a deal. However, I'd also really really like to see auto-detection of the common babel config files. I have a project that has a babel.config.js file _and_ a .babelrc file because of Mix. This would be a very welcome addition. Do you think you could cover that in this PR?

soilSpoon:
@thecrypticace Okay, I'll try it

soilSpoon:
I done. Ready to merge

thecrypticace:
This is awesome, thanks @soilSpoon! I'll be reviewing over the weekend.

src/BabelConfig.js
```diff
@@ -37,15 +78,7 @@ class BabelConfig {
...
-                    {
-                        modules: false,
-                        forceAllTransforms: true
-                    }
-                ]
-            ],
+            presets: ['@babel/preset-env'],
             plugins: [
                 '@babel/plugin-syntax-dynamic-import',
                 '@babel/plugin-proposal-object-rest-spread',
```
soilSpoon, line 84: It looks like this line is added optional. It may not be necessary for people using modern browsers.

And 85 to 90 lines

thecrypticace:
Hey @soilSpoon! Sorry for the rather long delay. This looks great! I've made some tweaks to handle babel.config.js files and some other things as well. There's still an issue related to `.babelrc` files I'm trying to resolve. I won't have more time to look into it for at least a few days most likely.

soilSpoon:
@thecrypticace  Can you tell if there are any issues with `.babelrc`?

thecrypticace:
I actually was able to fix that last night after I had a bit of an aha moment. There's an issue with the tests on windows though.

soilSpoon:
@thecrypticace Fixed the error. Is there anything more to do?

thecrypticace:
Much thanks @soilSpoon, this is great!

PR #2809, opened 2021-01-27
Fix  react-refresh in React Component

soilSpoon:
This is because when you enable overlay, you get websocket related errors.

I hope to re-enable after resolving this.

image

## Update
`@pmmmwh/react-refresh-webpack-plugin` was updated to the `^0.5.0-beta.0` version that supports `webpack-dev-server` `4.0.0-beta.0`, and solve the phenomenon that `hmr` does not work normally when using react components

### DIFF
--- a/src/Dependencies.js
+++ b/src/Dependencies.js
@@ -145,7 +145,7 @@ class Dependencies {
         }
 
         function isValid() {
-            return dep.check ? dep.check(require(name)) : true;
+            return dep.check ? dep.check(name) : true;
         }
 
         return {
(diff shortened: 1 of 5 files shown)


thecrypticace:
Could we figure out why the websocket url is invalid (I think it's missing the ws:// prefix) and fix that instead? If not then yeah we can merge this in.

soilSpoon:
I'll try it once on the weekend

soilSpoon:
@thecrypticace I fixed it!

soilSpoon:
Fixed it!

soilSpoon:
@JeffreyWay Could you please let me know when the changes in the PR will be reflected?

JeffreyWay:
@soilSpoon The plan is to push a tagged release out this weekend.

PR #2810, opened 2021-01-28
Update sass-loader

buberdds:
When sass-loader is missing in local package.json tap is installing an older version.


thecrypticace:
Heh, whoops. Thanks!

PR #2811, opened 2021-01-28
Set exit status 126 if warnings like "Please run Mix again." happens

dmitryuk:
Sometimes Laravel mix requires additional dependencies, which installs automatically with console message: "Please run Mix again."
The problem is that it exists with a default exit code=0, so it looks like command `npm run production` completed successfully and can be deployed to production (so it was already happened).
The PR set exit code to 126 "Command invoked cannot execute" and it can be catched by a CI correctly as fail.

### DIFF
--- a/src/Dependencies.js
+++ b/src/Dependencies.js
@@ -118,7 +118,7 @@ class Dependencies {
             );
 
             if (process.env.NODE_ENV !== 'test') {
-                process.exit();
+                process.exit(126);
             }
         }
     }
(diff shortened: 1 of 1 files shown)


src/Dependencies.js
```diff
@@ -118,7 +118,7 @@ class Dependencies {
             );
 
             if (process.env.NODE_ENV !== 'test') {
-                process.exit();
+                process.exit(126);
```
thecrypticace, line 121:
```suggestion
                process.exit(1);
```

126 isn't appropriate here as that is an indication that something is not a valid executable — or that you don't have permission to execute it.

thecrypticace:
Good catch! One minor thing to fix and we can merge this in.

dmitryuk:
Fixed

thecrypticace:
Thanks!

PR #2783, opened 2021-01-20
Recover https option in hot command

soilSpoon:
```
error: unknown option '--https'
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! working-note@ hot: `mix watch --hot --https`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the working-note@ hot script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```

I currently have the same problem as above on laravel-mix 6.
Fixes an issue where https is not properly supported when laravel-mix is upgraded to version 6.

PR #2835, opened 2021-02-10
HMR - makes laravel-mix do the right thing with a complex hot host

sharkey3:
makes laravel-mix do the right thing with a complex hot host setup such as

npx mix watch --hot -- --https --host=0.0.0.0 --port=8888 --public=myip.net

(I am developing on an AWS server with a load balancer frontend. The above command works, but laravel mix puts the wrong thing in public/hot. This makes it do the right thing and simplifies the code.)

### DIFF
--- a/src/HotReloading.js
+++ b/src/HotReloading.js
@@ -1,5 +1,6 @@
 let path = require('path');
 let File = require('./File');
+let argv = require('yargs').argv;
 
 module.exports = {
     record() {
@@ -10,7 +11,7 @@ module.exports = {
         }
 
         this.hotFile().write(
-            `${this.http()}://${Config.hmrOptions.host}:${this.port()}/`
+            `${this.http()}://${this.host()}:${this.port()}/`
         );
     },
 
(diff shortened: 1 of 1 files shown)


Verron:
@sharkey3 Does npx not throw an error when passing --https?

My output:

```
root@0561290c5aa9:/node# npx mix watch --hot --https
error: unknown option '--https'
```

I had to pass mix.webpackConfig({devServer: { https: true } }) in order to enable https.

I'm on webpack@5.14.0 and webpack-dev-serve@4.0.0-beta.0

thecrypticace:
@Verron The release adding the https option hasn't been tagged yet.

Verron:
@thecrypticace Thanks for clarifying.

sharkey3:
Please notice the "--" before "--https" 
That makes it an option for webpack, and not an option for mix.

buberdds:
Isn't https://github.com/JeffreyWay/laravel-mix/pull/2783/files resolves this issue? You can pass whatever you want via HMR options.

sharkey3:
pull 2783 does nothing with --host and --public
It is possible to do this with hmrOptions, if you spend the hours needed to figure out the right recipe, but this makes it do the right thing automatically.

damms005:
I'm not sure where to put this, but there is [a comment](https://github.com/laravel-mix/laravel-mix/issues/1721#issuecomment-418047997) that helps when dealing with non-default `host`.

Specifically, it suggested updating [the relevant doc](https://github.com/laravel-mix/laravel-mix/issues/1721#issuecomment-406767167) so users reading the HMR docs can know how to handle it.

However, the doc is still missing it. @JeffreyWay is this missing because you don't want to put too much in that section or do you welcome PR for this?

PR #2860, opened 2021-02-23
Reduce reliance on globals

thecrypticace:
The idea here is to reduce the use of implicit global `Mix` / `Config` / `webpackConfig` across the mix codebase. This will definitely be required if we ever want to support multiple / concurrent builds (a feature I'd absolutely love to have and have been slowly working towards support for). I've left in stubs in (hopefully) every place just in case plugins or user code are pulling in internal files.

**Future work**
I have not touched global usage within `src/components` as this will require some thought on the best way to expose the `Mix` object to components in a way that doesn't break things. Unfortunately we can't use the name `mix` because components already use that as a method to vend additional api. Maybe `context`?

I think `Context` could be a better name for `Mix` but we'd need to implement it with stubs for backwards compatibility.


thecrypticace:
@JeffreyWay I'd very much appreciate a code review here at some point if / when you get some time in the future.

JeffreyWay:
@thecrypticace Hey man. I have tomorrow set aside to work on Mix the full day, so I'll get back to you then. 👍

JeffreyWay:
Okay, this is a massive PR haha. I'd love it, if you can pull it off. I've had a lot of trouble in the past trying to figure out how to tackle it. 

Are you wanting to push this as a patch to Mix v6, or can we move it to a new v7 branch? I'm really worried about this breaking a bunch of unexpected things.

thecrypticace:
I was wanting to push this as a patch to Mix v6 and yeah I was worried about it breaking things which is why I left in a bunch of patches to keep any uses of globals inside external dependencies working.

I'm totally cool with pushing this to a v7 though if you want.

thecrypticace:
> Can we silently make this update, and then include a note in the release notes that global access is now deprecated and will trigger a warning once v7 is out?

Absolutely.

JeffreyWay:
Sounds great. I'll also tag a new release on Friday.

PR #2877, opened 2021-03-04
Fix mix.when()

claytonrcarter:
I just upgraded an app to mix 6.x and was getting errors for the following section of my `webpack.mix.js`:
```js
    .when(!mix.inProduction(), mix =>
        mix.copy('node_modules/vue/dist/vue.js', 'public/js')
    )
```
There error was `[webpack-cli] TypeError: mix.copy is not a function`, so it seems that the mix instance isn't being passed to the callback. Since [`mix.when()` is my baby](https://github.com/JeffreyWay/laravel-mix/pull/2330), I felt compelled to give it some love.

I haven't been following the development of mix 6 closely, but the move to the new component class system (in 5ad964f) appears to have broken `when()` b/c it changed what `this` refers to. Changing `this` to `Mix.api` appears to fix it, but I'm 100% open to corrections if this isn't the correct way to solve this.

While tinkering, I also realized that the return is no longer needed. I have added tests to verify these changes and I improved the name of the existing test for `mix.when()`.

Hope this helps! Thanks again. BTW the upgrade to v6 was silky smooth (aside from this hiccup).

### DIFF
--- a/test/features/when.js
+++ b/test/features/when.js
@@ -2,7 +2,7 @@ import test from 'ava';
 
 import '../helpers/mix';
 
-test('mix.when()', t => {
+test('it executes the callback based on the condition', t => {
     let called = false;
 
     mix.when(false, () => (called = true));
@@ -12,4 +12,15 @@ test('mix.when()', t => {
     mix.when(true, () => (called = true));
 
     t.true(called);
+});
+
+test('it passes the mix instance to the callback', t => {
+    mix.when(true, _mix => t.deepEqual(mix, _mix));
+});
+
+test('it returns the mix instance', t => {
+    t.deepEqual(
+        mix,
+        mix.when(true, () => {})
+    );
 });
--- a/test/features/when.js
+++ b/test/features/when.js
@@ -17,10 +17,3 @@ test('it executes the callback based on the condition', t => {
 test('it passes the mix instance to the callback', t => {
     mix.when(true, _mix => t.deepEqual(mix, _mix));
 });
-
-test('it returns the mix instance', t => {
-    t.deepEqual(
-        mix,
-        mix.when(true, () => {})
-    );
-});
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -46,8 +46,8 @@ declare namespace mix {
         /** Run a callback after the build has completed */
         then(callback: (stats: webpack.Stats) => void | Promise<void>): Api;
 
-        /** Run a callback after the build has completed */
-        when(callback: () => void | Promise<void>): Api;
+        /** Run a callback if a condition is true */
+        when(condition: boolean, callback: (mix: Api) => void | Promise<void>): Api;
     }
 
     // Webpack config related abilities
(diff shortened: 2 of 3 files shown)


test/features/when.js
```diff
@@ -13,3 +13,14 @@ test('mix.when()', t => {
...
+test('it passes the mix instance to the callback', t => {
+    mix.when(true, _mix => t.deepEqual(mix, _mix));
+});
+
+test('it returns the mix instance', t => {
+    t.deepEqual(
+        mix,
+        mix.when(true, () => {})
+    );
+});
```
thecrypticace, line 26: Lets remove this last test. Because of the way components work I think it's unnecessary. It doesn't matter what register() returns as its ignored and all mix components work this way.

thecrypticace (changes requested):

thecrypticace:
Oh, also would you please update the typescript types for mix.when()?

Make those two changes and this looks good to go!

test/features/when.js
claytonrcarter: No problem, I just made this change. FWIW, I copied that test more or less as-is from a similar test in `coffee.js`

https://github.com/JeffreyWay/laravel-mix/blob/f0ef8ab2370cb58fe656c9f285f8cb3144e38f40/test/features/coffee.js#L8-L10

claytonrcarter:
> Oh, also would you please update the typescript types for mix.when()?

:grimacing: Uh oh. I've never done typescript ... I just pushed an edit to `types/index.d.ts` in 6dd2e2a ... is that the right place to look, and does that edit look sorta/kinda OK? 

Sorry to need some hand holding on this...

types/index.d.ts
```diff
@@ -46,8 +46,8 @@ declare namespace mix {
         /** Run a callback after the build has completed */
         then(callback: (stats: webpack.Stats) => void | Promise<void>): Api;
 
-        /** Run a callback after the build has completed */
-        when(callback: () => void | Promise<void>): Api;
+        /** Run a callback if a condition is true */
+        when(condition: boolean, callback: (mix: Api) => void | Promise<void>): Api;
```
thecrypticace, line 50: welp I totally had the types wrong for .when. 😂 

This looks 100% correct, yes! Thank you!

claytonrcarter: *[Copied from another comment.]* Does that `| Promise<void>` look right to you? I don't think that when() would work if passed a Promise, but maybe I'm wrong.

thecrypticace:
> Sorry to need some hand holding on this...

That's quite alright. I'm more than happy to help out with this!

> Edit I'm not sure that | Promise<void> should be in the signature. Am I correct in that? I don't think that when() would work if passed a Promise, but maybe I'm wrong.

We're not awaiting it but it'll let users write `mix.when(async () => {})` without a type error. Maybe the better type is `unknown` or `any` but I think it's fine as is.

types/index.d.ts
thecrypticace: I'd leave it! We can always adjust the types again later if need be.

thecrypticace:
Thanks @claytonrcarter!

claytonrcarter:
Awesome. Thank you, Jordan. 🙌

PR #2883, opened 2021-03-09
Update HotReloading.js

asolopovas:
Leading slash produces the following hot http://localhost:8080/ that leads to a double slash like so in the link of dev server assets `http://localhost:8080//js/app.js`

### DIFF
--- a/src/HotReloading.js
+++ b/src/HotReloading.js
@@ -10,7 +10,7 @@ module.exports = {
         }
 
         this.hotFile().write(
-            `${this.http()}://${Config.hmrOptions.host}:${this.port()}/`
+            `${this.http()}://${Config.hmrOptions.host}:${this.port()}`
         );
     },
 


MonsieurMan:
Thank you, you saved me some more troubleshooting. This should be merged and released because it breaks the usage of hot reloading.

thecrypticace:
So, I hadn't merged this yet because I thought we'd need to coordinate a release with Laravel to normalize the `//` -> `/` mainly in the case of an asset not beginning with a slash causing an invalid url. However, poking at the source code now it fixes it up if something doesn't already (and has done so almost since the inception of the mix() helper)

What's weid is this shouldn't cause an issue anyway because I worked around the problem with this: https://github.com/JeffreyWay/laravel-mix/blob/3fa559471aa0bb092934f8dc72601f6bd144cbec/src/builder/WebpackConfig.js#L136-L142

I use HMR a decent amount and it's been working so it's possible a recent change to a dependency may have broken it.

thecrypticace:
Thanks!

Issue #2340, opened 2020-03-06
css-preprocessors documentation out of date


jakewhiteley-gc:
-   Laravel Mix Version: 5.0.1
-   Node Version: 11.9.0
-   NPM Version: 6.14.2
-   OS: Windows

### Description:
The docs for the 5.* branch are out of date.

At the URL [here](https://laravel-mix.com/docs/5.0/css-preprocessors) the site states to use the following code:

``` js
mix.sass('resources/sass/app.scss', 'public/css')
   .options({
        autoprefixer: {
            options: {
                browsers: [
                    'last 6 versions',
                ]
            }
        }
   });
```

Which produces the following console error:

```bash 
You did not set any plugins, parser, or stringifier. Right now, PostCSS does nothing.
 Pick plugins for your case on https://www.postcss.parts/ and use them in postcss.con
fig.js.
```

As mentioned in issue #2273 , the `enabled: true` option needs to be redeclared like so:

```js
    .options({
        autoprefixer: {
            enabled: true,
            options: {
                browsers: [
                    'last 2 versions',
                ]
            }
        }
    })
```

jakewhiteley-gc:
bumping

PR #2894, opened 2021-03-16
Update autoprefixer.md

ethanclevenger91:
I couldn't get autoprefixer to config properly with the documented method. Did this instead, worked. Can't vouch for whether or not the existing documentation works as advertised for postCss directly so I left it intact.

### DIFF
--- a/docs/autoprefixer.md
+++ b/docs/autoprefixer.md
@@ -74,6 +74,17 @@ mix.postCss('src/app.css', 'dist')
 If you instead wish to disable autoprefixing entirely, set `autoprefixer` to `false`.
 
 ```js
+// Using Sass
+mix.sass('src/app.css', 'dist')
+   .options({
+      postCss: [
+         require('autoprefixer')({
+            remove: false,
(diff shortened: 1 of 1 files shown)


thecrypticace:
If `{ autoprefixer: false }` isn't working we should definitely figure out how to fix that instead because that should work.

ethanclevenger91:
I was actually trying to tweak autoprefixer options. I suppose it's possible that disabling it entirely is simpler than setting options?

```
.sass('app.scss', 'dist')
    .options({
        postCss: [
            require('autoprefixer')({
                grid: true,
            })
        ]
    })
```

ethanclevenger91:
@thecrypticace I guess my issue could have actually been [this](https://github.com/JeffreyWay/laravel-mix/issues/2340#issue-577126665)

thecrypticace:
I've verified that, at least in the latest version of Mix, this is not an issue and the autoprefixer config works as expected when using postcss or a preprocessor like sass/scss

Issue #2910, opened 2021-03-24
Image extensions case + collisions in image names when the same image used across the app


aarbi:
https://github.com/JeffreyWay/laravel-mix/blob/798424b45cd668ddc3c289ecb20f28797b23a748/src/builder/webpack-rules.js#L21

```
[webpack-cli] Error: Prevent writing to file that only differs in casing or query string from already written file.
This will lead to a race-condition and corrupted files on case-insensitive file systems.
/Users/batcave/laravel/public/images/12.png
/Users/batcave/laravel/public/images/12.png
```

I suggest two things:

1. make the regex case insensetive ( because there are some files with `.PNG` extension
2. prefix the output name with hash to avoid collisions

aarbi:
Check the linked PR for details

PR #2911, opened 2021-03-24
GH-2910: Make the test for images case insensitive

aarbi:
fixes GH-2910


JeffreyWay:
I can make the regex case sensitive, but I'm pretty sure this is a potentially breaking change. Some people will be hardcoding their asset paths, and it'll break due to the hash being added to the file name.

aarbi:
Makes sense, going to close it

PR #2930, opened 2021-04-08
Change default BrowserSync Proxy to folder name

chrisrenga:
changes the default proxy to the project root folder name followed by ".test". Inspired by Laravel Vale's domain naming convetion.

before

```js
let path = require("path");

...

mix.broserSync({
    proxy: path.basename(__dirname) + '.test',
})
```

after

```js
mix.browserSync()
```

### DIFF
--- a/src/components/BrowserSync.js
+++ b/src/components/BrowserSync.js
@@ -43,7 +43,7 @@ class BrowserSync {
         let defaultConfig = {
             host: 'localhost',
             port: 3000,
-            proxy: process.cwd().split('/').pop() + '.test',
+            proxy: process.cwd().replaceAll('\\', '/').split('/').pop() + '.test',
             files: [
                 'app/**/*.php',
                 'resources/views/**/*.php',
--- a/src/components/BrowserSync.js
+++ b/src/components/BrowserSync.js
@@ -1,3 +1,5 @@
+let path = require('path');
+
 class BrowserSync {
     /**
      * Required dependencies for the component.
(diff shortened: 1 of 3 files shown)


src/components/BrowserSync.js
```diff
@@ -43,7 +43,7 @@ class BrowserSync {
         let defaultConfig = {
             host: 'localhost',
             port: 3000,
-            proxy: 'app.test',
+            proxy: process.cwd().replace(/\\/g, '/').split('/').pop() + '.test',
```
thecrypticace, line 46: This would be a bit more concise:

```suggestion
            proxy: `${path.basename(process.cwd())}.test`,
```

thecrypticace: We'd want to require path at the top as well.

test/features/browsersync.js
```diff
@@ -45,7 +45,7 @@ test('it injects the snippet in the right place', t => {
 });
 
 test('it configures Browsersync proxy', t => {
-    t.is(browserSyncConfig().proxy, 'app.test', 'sets default proxy');
+    t.is(browserSyncConfig().proxy, 'laravel-mix.test', 'sets default proxy');
```
thecrypticace, line 48: This test would fail if you don't name the laravel-mix repo folder `laravel-mix` which feels flaky to me. I don't know a good solution to that though.

src/components/BrowserSync.js
chrisrenga: Thanks

test/features/browsersync.js
chrisrenga: will just do the same and set it to the current folder name. Thanks

PR #2939, opened 2021-04-19
Extract CSS for a react project

aarbi:
Using the same API as Vue, we can pass an options object that has `extractStyles` key to enable the css extract.

This also can be passed as global option to `mix.options`, but not implemented in this PR

PS: This will generate the chunks directly under `public/` folder


thecrypticace:
Good idea!

There's a few things we'll want to update here:
1. Ensure we update the split chunks list to pull rendered CSS out of React JSX files (Example: src/components/Vue.js @ updateChunks)
2. Add doc blocks for register() (Example: src/components/Vue.js @ register)
3. Update the TypeScript typings for react()
4. Add tests for these (we have vue tests for extractStyles — you should be able to mostly copy those)

I'd be more than happy to help out where needed!

aarbi:
@thecrypticace, thanks for the review, great points.
I took care of 2 and 3. I will get back to 1 and 4 in the upcoming days. Feel free to jump in!

aarbi:
I tried to replicate the logic of updateChunks and other methods it needs, I ended up with the following exported files
```
/js/app.css
/js/app.css.map
/js/app.js
/js/app.js.LICENSE.txt
/js/app.js.map
176.css
176.css.map
195.css
195.css.map
229.css
229.css.map
321.css
321.css.map
386.css
386.css.map
459.css
459.css.map
531.css
531.css.map
593.css
593.css.map
625.css
625.css.map
785.css
785.css.map
903.css
903.css.map
966.css
966.css.map
css/index.css
css/index.css.map
```

my current `webpack.mix.js` is the following:
```
mix
  .js('resources/js/app.jsx', 'public/js/app.js')
  .react({
    extractStyles: mix.inProduction()
  })
  .sourceMaps()
  .sass('resources/scss/index.scss', 'public/css/index.css');
```
The `/js/app.css` looks a bit weird, but it represents the css/sass loaded in app.jsx. Yet that should ideally go to `public/css`.
I would need some help to fix that

mikkpokk:
@thecrypticace @aarbi 

Below is the code I am using as a mix module at the moment. It also supports specifying `localIdentName` to css-loader. Tested with `Laravel Mix 6.0.34`.

It would be great to see it as built-in functionality in the future.

```javascript
const {Chunks} = require('laravel-mix/src/Chunks')
const File = require('laravel-mix/src/File')

class React {
    /**
     * Create a new component instance.
     */
    constructor() {
        this.chunks = Chunks._instance

        if (!this.chunks)
            this.chunks = new Chunks(Mix)
    }

    /**
     * Required dependencies for the component.
     */
    dependencies() {
        const dependencies = ['@babel/preset-react']

        if (this.supportsFastRefreshing()) {
            return dependencies.concat([
                {
                    package: '@pmmmwh/react-refresh-webpack-plugin@^0.5.0-rc.0',
                    check: name =>
                        semver.satisfies(
                            require(`${name}/package.json`).version,
                            '^0.5.0-rc.0',
                        ),
                },
                'react-refresh',
            ])
        }

        return dependencies
    }

    /**
     * Override the generated webpack configuration.
     *
     * @param {Object} config
     */
    webpackConfig(config) {
        // Loop through all rules
        config.module.rules = config.module.rules.map((rule) => {
            if (!rule.use && !rule.oneOf) {
                return rule;
            }

            // Loop through all loaders on regular use options
            rule.use = Array.isArray(rule.use) ? rule.use.map(this.replaceLoaderOptions) : rule.use;

            // Alternatively, loop through all of the oneOf options (if they exist)
            rule.oneOf = Array.isArray(rule.oneOf)
                // And then replace all the loaders on each individual option
                ? rule.oneOf.map(oneOf => {
                    oneOf.use = oneOf.use.map(this.replaceLoaderOptions);
                    return oneOf;
                })
                : rule.oneOf;

            return rule;
        });

        this.updateChunks()

        return config
    }

    /**
     * This function replaces the loader options for any css-loader rules that are
     * found in the webpack config.
     * @param {*} rule
     */
    replaceLoaderOptions = (rule) => {
        // If the loader is not the css-loader, we can safely skip it
        if (rule.loader.indexOf('/css-loader/') === -1 && rule.loader !== 'css-loader' && rule !== "css-loader") {
            return rule;
        }

        if (!rule?.options?.modules)
            return rule;

        // Add our options to the loader
        let options = {
            ...(rule?.options ?? {}),
            modules: {
                ...(rule?.options?.modules ?? {}),
                mode: "local",
                localIdentName: this.options.localIdentName,
            },
        };

        // Convert string syntax to object syntax if neccessary
        rule =
            typeof rule === "string"
                ? {
                    rule,
                }
                : rule;

        // Inject our options into the loader
        rule.options = rule.options
            ? Object.assign({}, rule.options, options)
            : options;

        return rule;
    }

    /**
     * Register the component.
     *
     * @param {object} options
     * @param {boolean|string} [options.extractStyles] Whether or not to extract React styles. If given a string the name of the file to extract to.
     */
    register(options = {}) {
        if (
            arguments.length === 2 &&
            typeof arguments[0] === 'string' &&
            typeof arguments[1] === 'string'
        ) {
            throw new Error(
                'mix.react() is now a feature flag. Use mix.js(source, destination).react() instead',
            )
        }

        this.options = Object.assign(
            {
                extractStyles: false,
                localIdentName: '[name]_[local]__[hash:base64:5]',
            },
            options,
        )

        Mix.extractingStyles = !!this.options.extractStyles
    }

    /**
     * webpack plugins to be appended to the master config.
     */
    webpackPlugins() {
        if (!this.supportsFastRefreshing()) {
            return []
        }

        const ReactRefreshPlugin = require('@pmmmwh/react-refresh-webpack-plugin')

        return new ReactRefreshPlugin({overlay: {sockPath: 'ws'}})
    }

    /**
     * Babel config to be merged with Mix's defaults.
     */
    babelConfig() {
        const plugins = this.supportsFastRefreshing()
            ? [require.resolve('react-refresh/babel')]
            : []

        return {
            presets: [['@babel/preset-react', {runtime: 'automatic'}]],
            plugins,
        }
    }

    /**
     * Determine if the React version supports fast refreshing.
     */
    supportsFastRefreshing() {
        return Mix.isHot() && semver.satisfies(this.library().version, '>=16.9.0')
    }

    /**
     * Load the currently installed React library.
     */
    library() {
        return require('react')
    }

    /**
     * Update CSS chunks to extract React styles
     */
    updateChunks() {
        if (this.options.extractStyles === false) {
            return
        }

        this.chunks.add(
            'styles-js',
            this.styleChunkName(),
            [/.(j|t)s$/, module => module.type === 'css/mini-extract'],
            {
                chunks: 'all',
                enforce: true,
                type: 'css/mini-extract',
            },
        )

        this.chunks.add(
            'styles-jsx',
            this.styleChunkName(),
            [/.(j|t)sx?$/, module => module.type === 'css/mini-extract'],
            {
                chunks: 'all',
                enforce: true,
                type: 'css/mini-extract',
            },
        )
    }

    /**
     * Get the name of the style chunk.
     *
     * @returns {string}
     */
    styleChunkName() {
        // If the user set extractStyles: true, we'll try
        // to append the React styles to an existing CSS chunk.
        if (this.options.extractStyles === true) {
            let chunk = this.chunks.find((chunk, id) => id.startsWith('styles-'))

            if (chunk) {
                return chunk.name
            }
        }

        return this.extractFile().relativePathWithoutExtension()
    }

    /**
     * Get a new File instance for the extracted file.
     *
     * @returns {File}
     */
    extractFile() {
        return new File(this.extractFileName())
    }

    /**
     * Determine the extract file name.
     *
     * @return {string}
     */
    extractFileName() {
        let fileName =
            typeof this.options.extractStyles === 'string'
                ? this.options.extractStyles
                : '/css/app-styles.css'

        return fileName.replace(Config.publicPath, '').replace(/^\//, '')
    }
}
```

mikkpokk:
@aarbi Please see my shared code above, it also includes possibility to adjust localIdentName which is quite important to a React developers on debugging purposes.

There are some missing parts against latest Mix React.js component also: https://github.com/laravel-mix/laravel-mix/blob/master/src/components/React.js

aarbi:
I remerged master again.
@mikkpokk push the changes you proposed to https://github.com/aarbi/laravel-mix/tree/patch-2 so that your contribution remains in it!

mikkpokk:
@aarbi Done, you forgot to call updateChunks method also.

PR pushed into your repository https://github.com/aarbi/laravel-mix/pull/1

mikkpokk:
Hello @thecrypticace 

Please review this pull request. The PR is solving issues #2940 and #3123.

mikkpokk:
@thecrypticace Ready to merge from our side.

thecrypticace:
I cleaned up the code a bit and made css modules configurable globally via mix.options: `mix.options({ cssModuleIdentifier: '[name]_[local]__[hash:base64:5]' })`

Couple of questions:
1. I kept the default set to `[hash:base64]`. Should calling mix.react() override the default or do you think it's fine to leave the default as is?
2. What's the reason for adding `skipEnvCheck: true` to `react-refresh/babel`? Should we just opt to not register that in production mode instead?
3. Is `cssModuleIdentifier` a reasonable option name? I didn't use localIdentName because that's css-loader specific and could change (also it's name doesn't imply that it has to do anything with css).

aarbi:
For question 2:
 I got that error while running the tests, since babel detects that react-refresh plugin is loaded in non dev environment, `test` env to be specific

aarbi:
Here is an example of the issue I got for question 2: https://github.com/laravel-mix/laravel-mix/runs/4130794906?check_suite_focus=true I don't know if there is any better solution to overcome this issue
Also, for the other two questions, both of the defaults you introduced are  good!

thecrypticace:
Thanks! 🎉

PR #2946, opened 2021-04-22
[#2945] Improve Yarn Compatibility 

### LINKED ISSUE #2945: `npx` causing issues with Yarn
-   Laravel Mix Version: ^6.0.18 (`npm list --depth=0`)
-   Node Version (`node -v`): v14.16.0
-   NPM Version (`npm -v`): 6.14.11
-   Yarn Version: 2.x and 3.0.0-rc.2
-   OS: macOS 11.2.3

### Description:

The `watch` and `build` commands run `npx` under the hood. This causes issues when using Yarn as the package manager and forces users to install `webpack` and `webpack-cli` as dependencies manually resulting in a confusing experience.

### Steps To Reproduce:

- Create a project using Yarn as the package manager
- Add `laravel-mix` as a dev dependency
- Add a `webpack.mix` file (or compatible variant)
  - This only needs to contain the import at the top of the file (`const mix = require('laravel-mix');`)
- Run `mix` using Yarn — `yarn mix`

On first run, the user is prompted to install `webpack-cli`.  If the user chooses to install at that time, the command then subsequently fails as it can't find `webpack`. In order to fix this issue, `webpack` must also be installed as a dev dependency.

### Solution:

Rather than running these commands with `npx`, `build` and `watch` should call `webpack` directly and both `webpack` and `webpack-cli` should be listed as peer dependencies.

zulaica:
Closes #2945 

- Moves `webpack` and `webpack-cli` to list of peer dependencies
- Updates commands to call `webpack` directly

### DIFF
--- a/types/component.d.ts
+++ b/types/component.d.ts
@@ -1,7 +1,8 @@
 // Type definitions for laravel-mix 6.0
 
 import * as webpack from 'webpack';
-import { TransformOptions as BabelConfig } from 'babel-core';
+import { TransformOptions as BabelConfig } from '@babel/core';
+import api from './index';
 
 export type DependencyObject = {
     /** The name of the package */
@@ -73,7 +74,7 @@ export interface ClassComponent {
 }
 
 export interface FunctionalComponent {
-    (mix: Api, config: webpack.Configuration, ...args: any[]): void;
+    (mix: typeof api, config: webpack.Configuration, ...args: any[]): void;
 }
 
 export type Component = ClassComponent | FunctionalComponent;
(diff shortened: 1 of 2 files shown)


thecrypticace:
I feel like this could break things. peer dependencies are **not** installed by default. They're meant to be provided by the root installed package.

zulaica:
@thecrypticace That's true, but `postcss` is already a peer dependency, so I figured users would already be fine with adding a couple more.

That said, I've set this back as a draft while I work on another solution that I've referenced in the linked issue. Thanks for taking a look!

thecrypticace:
Postcss is a peer dependency and we handle installation for you when you first run Mix if it's not installed. We _could_ do the same with webpack but that seems to defeat the purpose since mix is a wrapper around webpack. If we wanted to such a change couldn't happen until Mix v7 at the earliest.

zulaica:
Oh, it does! Yarn 2+ issues a warning on first `yarn install` that the `postcss` is missing.

```
➤ YN0002: │ example@workspace:. doesn't provide postcss (pf3a0b), requested by laravel-mix
```

So, of course, I went ahead and installed it manually before calling `yarn mix`. But I just tried it right now without installing `postcss` before running `yarn mix` and sure enough `mix` installed it for me. And, yeah, that does complicate things. I'll keep poking around to see what potential solutions there might be. And thanks for the input! You've been very helpful.

PR #2965, opened 2021-05-07
Pin to previous webpack-dev-server release

thecrypticace:
The original plan was to upgrade but this is the safer option until I've had time to investigate the changes.


thecrypticace:
ping @JeffreyWay could we get a tag today if you've got time?

PR #2972, opened 2021-05-11
Allow API to be extended by third party plugins

markhalliwell:
Currently, due to namespacing and not exporting a default, extending `Api` to allow third party plugins to add to it isn't possible:

### Before (namespaced):
- Plugin's `index.d.ts` definition:
  namspaced-definition

- Autocomplete:
  current-autocomplete

### After (not namespaced):
- Plugin's `index.d.ts` definition:
  no-namespaced-definition

- Autocomplete:
  new-autocomplete

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -20,271 +20,269 @@ import { VueLoaderOptions } from 'vue-loader';
 import MixHelpers from '../src/Mix';
 
 // General API
-declare namespace mix {
-    export interface Api {
-        /** Set the public path */
-        setPublicPath(path: string): Api;
+export interface Api {
+    /** Set the public path */
+    setPublicPath(path: string): Api;
 
-        /** Set the resource path */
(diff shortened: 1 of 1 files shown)


thecrypticace (approved):

thecrypticace:
Alright I'm getting back to this finally. `export default` does indeed change the behavior but for whatever reason it didn't when I tested before. I'm trying to come up with a solution but failing at the moment. Every solution I find requires using ambient modules & namespaces.

Any ideas?

thecrypticace:
Alright I think I figured it out after some experimentation. I had to
1. Change all `export interface` to `interface`
2. Remove the default export
3. Add an export assignment like this:

```
declare const exports: Api;
declare namespace exports {
    export { Api, VueConfig };
}

export = exports;
```

I got the idea from looking at the webpack types. It seems to work as far as I can tell.

PR #2988, opened 2021-05-25
Fix already defined error messages

lcoclea:
This PR is to fix already defined property error.

How to replicate:

- create an npm package and add laravel-mix as a dependency
- create a second npm package and add the first package as a dependency and laravel-mix


thecrypticace:
I'd love to just get rid of these but I don't want to possibly break other packages in a minor / patch release. Thanks!

Issue #1053, opened 2017-07-26
BrowserSync: Sass compilation causing whole page reload


SethTompkins:
- Laravel Mix Version: 1.4.2 (github:jeffreyway/laravel-mix#5aad0dfde5349e284e0876388a6f307547f7c002)
- Node Version (`node -v`): 8.2.1
- NPM Version (`npm -v`): 5.3.0
- OS: macOS Sierra 10.12.6 

### Description:
When using browserSync, css changes are injected however the page is also reloaded due to the fact that the JS bundle is re-emitted even when the only file that changed was a sass file.

### Steps To Reproduce:

webpack.mix.js:
```
mix
	.js('resources/assets/js/main.js', 'public/assets/js/bundle.js')
	.extract(['jquery', 'babel-polyfill'])
	.autoload({
		jquery: ['$', 'jQuery', 'window.jQuery']
	})
	.sass('resources/assets/sass/style.scss', 'public/assets/css')
	.disableNotifications()
	.options({
			processCssUrls: false
	})
	.sourceMaps()
	.browserSync({
		proxy: process.env.APP_URL,
        files: [
            'app/**/*.php',
            'resources/views/**/*.php',
            'public/assets/js/**/*.js',
            'public/assets/css/**/*.css'
        ]
    })
```

Run `npm run watch`, site is served at localhost:3000, save a sass file and watch the browser refresh rather than just inject the changes.

marcamos:
My setup is similar to yours, save for defining exactly which .css and .js file to 'watch' for — see the first two entries in the `files: []` array:
```
mix.autoload({jquery: ['$', 'window.jQuery', 'jQuery']})
   .js('_js/app.js', 'public/lib/js')
   .sass('_scss/app.scss', 'public/lib/css')
   .options({
     processCssUrls: false
   })
   .sourceMaps()
   .browserSync({
     proxy: 'starter.dev',
     files: [
       'public/lib/css/app.css',  // Generated .css file
       'public/lib/js/app.js',    // Generated .js file
       'public/**/*.+(html|php)',           // Generic .html and/or .php files [no specific platform]
       'laravel/resources/views/**/*.php',  // Laravel-specific view files
       'craft/templates/**/*.+(html|twig)'  // Craft-specific templates, as html and/or twig
     ]
   });
```
Perhaps that has something to do with it?

SethTompkins:
I tried this, and it did not work for me:

```
.browserSync({
[...]

PR #2991, opened 2021-05-28
Only build one time in CustomTasksPlugin (#1053)

cgodo:
This solves an issue with unnecessary reloads with BrowserSync: https://github.com/JeffreyWay/laravel-mix/issues/1053#issuecomment-826368904

`CustomTasksPlugin` runs all post-build tasks again for every rebuild, no matter if it was not initiated from its own watchers. All files built using `.babel()`, `.scripts()`, `.combine()`, `.styles()` and `.copy()` are being rebuilt every time, causing unnecessary BrowserSync full page reloads.

For example, if you have a Vue app built with `.js()`, but also some old scripts built with `.babel()`. When changing something in the Vue app, it should apply a Hot Module Replacement without reloading the page, but since the `.js` files built with `.babel()` are also rebuilt, BrowserSync detects those changes and triggers a reload.

With this PR, CustomTasksPlugin only builds files the first time or when their watchers are triggered.

### DIFF
--- a/src/webpackPlugins/CustomTasksPlugin.js
+++ b/src/webpackPlugins/CustomTasksPlugin.js
@@ -17,27 +17,23 @@ class CustomTasksPlugin {
      * @param {import("webpack").Compiler} compiler
      */
     apply(compiler) {
-        let firstRun = true;
-        compiler.hooks.done.tapAsync(this.constructor.name, (stats, callback) => {
-            // Only run all tasks on first run
-            if (firstRun) {
-                firstRun = false;
-                this.runTasks(stats).then(async () => {
-                    await this.afterChange();
+        let ranOnce = false;
+        compiler.hooks.done.tapPromise(this.constructor.name, async stats => {
+            // Only run all tasks once
+            if (ranOnce) return;
+            ranOnce = true;
 
-                    if (this.mix.isWatching()) {
-                        this.mix.tasks.forEach(task =>
-                            task.watch(
-                                this.mix.isPolling(),
-                                debounce(this.afterChange.bind(this), 100)
-                            )
-                        );
-                    }
+            await this.runTasks(stats);
 
-                    callback();
-                });
-            } else {
-                callback();
+            await this.afterChange();
+
+            if (this.mix.isWatching()) {
+                this.mix.tasks.forEach(task =>
+                    task.watch(
+                        this.mix.isPolling(),
+                        debounce(this.afterChange.bind(this), 100)
+                    )
+                );
             }
         });
     }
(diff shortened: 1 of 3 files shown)


thecrypticace:
I'm fairly certain this will break builds that compile JS and then concatenate it afterward using `mix.scripts()`

cgodo:
> I'm fairly certain this will break builds that compile JS and then concatenate it afterward using `mix.scripts()`

But then it would trigger the watchers of those scripts, wouldn’t it? The final scripts would be concatenated again if they changed.

The changes proposed still maintain the watchers active. They are subscribed on the first run.

cgodo:
@thecrypticace , now while tasks only run on the first build, any updates on their files being watched will trigger the task and ALSO update the manifest file and minification (in case you were watching with production mode).

I tested this with a build that compiles JS files and then concatenates all afterward using `mix.scripts()`. It should work as usual now.

cgodo:
Also added a debounce to the `afterChange` function when executed from a watch change in a task. Versioning and minification perform better this way for cases like using `copy()` with lots of js files in watch mode.

src/webpackPlugins/CustomTasksPlugin.js
```diff
@@ -16,23 +17,28 @@ class CustomTasksPlugin {
      * @param {import("webpack").Compiler} compiler
      */
     apply(compiler) {
+        let firstRun = true;
         compiler.hooks.done.tapAsync(this.constructor.name, (stats, callback) => {
```
thecrypticace, line 21: We should change this to tapPromise and make this an async/await function. That'd clean things up a good bit here.

src/webpackPlugins/CustomTasksPlugin.js
```diff
@@ -16,23 +17,28 @@ class CustomTasksPlugin {
...
-                    await this.minifyAssets();
-                }
-
-                if (this.mix.isWatching()) {
-                    this.mix.tasks.forEach(task => task.watch(this.mix.isPolling()));
-                }
-
-                this.mix.manifest.refresh();
+            // Only run all tasks on first run
+            if (firstRun) {
```
thecrypticace, line 23: Can we flip the conditional to reduce the nesting?

thecrypticace:
I still need to do some testing with this just not sure when I'll get the time but I've left a few comments. I honestly think that everything driving custom tasks would be better suited as proper webpack plugins that add assets and what not but that would be a major breaking change (and also possibly quite difficult).

cgodo:
Added the fixes. Thanks for the comments.

thecrypticace:
Pulling this into v6.1. Thanks! I'll need to do more testing and tweaking most likely. Tasks aren't running when they need to be at the moment anyway.

PR #3015, opened 2021-06-17
Audit fixes

thecrypticace:
I've inlined the Browsersync types so we're not pulling in a vulnerable version of chokidar. This should leave one last vulnerability. We have to wait either for svgo to update for that fix.


jsandfordhughes:
@JeffreyWay it seems like the upgrade to `"webpack-dev-server": "4.0.0-beta.3"` has broken something with the websocket connection of the devServer. If your dev server is running in HTTPS mode the websocket connection no longer uses WSS and SSL which browsers block.

thecrypticace:
Can you open a separate issue for this?

PR #3018, opened 2021-06-19
Don't double load postcss.config.js

thecrypticace:
postcss-loader already does this on its own. If we pass in a list it can cause duplicated plugins inside postcss-loader even after it does plugin de-duplication due to reliance on stable object identity.

Fixes https://github.com/tailwindlabs/tailwindcss/issues/4712

### DIFF
--- a/src/PostCssPluginsFactory.js
+++ b/src/PostCssPluginsFactory.js
@@ -1,5 +1,3 @@
-let postcssrc = require('postcss-load-config');
-
 class PostCssPluginsFactory {
     /**
      * Create a new instance.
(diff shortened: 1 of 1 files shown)


thecrypticace:
@JeffreyWay could we get a tag when you get a chance?

This was a fun one to track down 😂

DarkGhostHunter:
Can confirm fixed using [latest master](https://github.com/JeffreyWay/laravel-mix/commit/6fdf6dbb2e0a9bc67350a2e57a0c3284b570f060).

PR #3050, opened 2021-07-26
Don’t force require dependencies

thecrypticace:
We definitely need a way run this scenario in the test suite.

I haven't tested this myself yet. I will later today when I get some time to. @JeffreyWay A code review on this would be helpful. _sigh_ me and late nights aren't mixing well lately.

### DIFF
--- a/src/Resolver.js
+++ b/src/Resolver.js
@@ -13,7 +13,11 @@ class Resolver {
             return this.aliases[name];
         }
 
-        return require.resolve(name);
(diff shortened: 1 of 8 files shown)


JeffreyWay:
Looks okay to me. I hate that this is even required in the first place. I'll tag a patch real quick.

thecrypticace:
Thanks, @JeffreyWay. I appreciate it!

Issue #3049, opened 2021-07-26
Building assets in watch mode does not re-build the file after initial change


jcordingley:
-   Laravel Mix Version: 6.0.25
-   Node Version (`node -v`): 14.7.3
-   NPM Version (`npm -v`):7.19.1
-   OS: Windows 10

### Description:
When running `npx mix watch` or `npx mix watch -- --watch-options-poll=1000` laravel-mix builds the assets for the first change but after a subsequent changes it says it has built the assets but the file does not change.

### Steps To Reproduce:

webpack.mix.js
```
const mix = require('laravel-mix');
const path = require('path');
require('laravel-mix-clean');

const entryPoints = {
    '/js/mixer' : 'Components/Mixer'
};

mix.js('resources/js/app.js', 'js')
    .setPublicPath('public')
    .webpackConfig({
        entry: entryPoints,
        resolve: {
            alias: {
                EntryPoints: path.resolve(__dirname, 'resources/js'),
                Libraries: path.resolve(__dirname, 'resources/js/lib'),
                Components: path.resolve(__dirname, 'resources/js/components'),
                Mixins: path.resolve(__dirname, 'resources/js/mixins'),
                Stores: path.resolve(__dirname, 'resources/js/stores')
            }
        },
    })
    .vue()
    .version()
    .clean({
        paths : [
            path.resolve(__dirname, 'public/css/**/*.css'),
            path.resolve(__dirname, 'public/js/**/*.js'),
            path.resolve(__dirname, 'public/js/*.js')
        ]
    });
```

When making changes to components / vuex stores etc after the first build then no changes are made to any build file

PR #3071, opened 2021-08-20
[Fix] watch-poll command detection

VitorHFLopes:
# Problem:

The new `mix watch -- --watch-options-poll=1000` command is not starting the watch with the polling option.

In Laravel Mix v6, the watch command as changed from `--watch-poll` to `--watch-options-poll={POLLING_INTERVAL}` and the current validation to test if `Mix` should use the polling options isn’t really detecting it.

# How to reproduce:

Just try to run the new default mix watch poll execution - `npm run watch-poll` or `npx mix watch -- --watch-options-poll=1000` - then change a file that has been “watched”.

# Solution:

As the command changed to receive the polling interval direct from the command, the current validation (`process.argv.includes('--watch-poll')`) actually doesn’t detect the option, because the `.includes()` method from JS Array needs to test exactly the indexes and the current validation should be prepared to test all the values passed by the developer.

Therefore, to simplify, I loop through all the `process.argv` values passed by the Node execution to test if one of them has the `—watch-options-poll` “key”.

Found by my teammate: [Irineu Junior](https://github.com/irineujunior)

This PR maybe closes the issue [#3049](https://github.com/JeffreyWay/laravel-mix/issues/3049)

### DIFF
--- a/src/Mix.js
+++ b/src/Mix.js
@@ -182,7 +182,9 @@ class Mix {
      * Determine if polling is used for file watching
      */
     isPolling() {
-        return this.isWatching() && process.argv.includes('--watch-poll');
+        const hasPollingOption = process.argv.some((arg) => arg.includes('--watch-options-poll'));
+
+        return this.isWatching() && hasPollingOption;
     }
 
     /**


JeffreyWay:
Thanks!

PR #3095, opened 2021-09-10
Allow changing the manifest name / path

thecrypticace:
It is now possible to customize the name / path to the mix manifest. Additionally, if you do not want a manifest to be generated it may be disabled.

If you want to rename your file to `manifest.json` you can change the name by setting the `manifest` option:

```js
mix.options({ manifest: 'manifest.json' })
```

This also accepts relative paths which means the following are possible:

```js
mix.options({ manifest: 'build/mix-manifest.json' }) // ./public/build/mix-manifest.json
mix.options({ manifest: '../mix-manifest.json' }) // ./mix-manifest.json
```

Additionally, if you don't want to generate the manifest at all you can do so by setting the option to `false`:
```js
mix.options({ manifest: false })
```

cc @sbine

### DIFF
--- a/src/Manifest.js
+++ b/src/Manifest.js
@@ -2,6 +2,8 @@ let collect = require('collect.js');
 let path = require('path');
 let File = require('./File');
 
+let hasWarnedAboutManifestChange = false;
+
 class Manifest {
     /**
      * Create a new Manifest instance.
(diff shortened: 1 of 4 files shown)


thecrypticace:
@sbine If this looks usable for you I can merge & tag v6.0.31 today

sbine:
@thecrypticace looks perfect and that was super fast, thanks!

thecrypticace:
Test failure is unrelated. Just because the integration tests on windows are slow. Merging…

thecrypticace:
Tagged & Published

PR #3096, opened 2021-09-11
WIP: Multiple build support

thecrypticace:
# This is a work in progress — much is still broken and/or all features documented or otherwise are subject to change

I'm opening this PR now to solicit feedback from any passersby about features and needed functionality.

--- 

# Support for multiple builds via "Build Groups"

Build groups are separate named configurations inside a single webpack.mix.js file. A build group can define it's own assets, behavior, plugins, webpack config, etc. The builds of which are processed independently.

## Defining build groups

Defining a build group requires calling `mix.group` with two arguments: a name and a callback. Providing a name allows you to tell Mix to only build a specific group (or set of groups). Build group names are case insensitive. It is an error to define the same group multiple times. You may use any name you like for a group including spaces but we recommend slug-like names separated by hyphens or underscores.

The example below defines two separate builds for app1 and app2 that can be built independent of one another.

```js
mix.group("app1", (mix) => {
    mix.js("src/app1/js/index.js", "dist/app1.js")
    mix.css("src/app1/css/index.css", "dist/app1.css")
})

mix.group("app2", (mix) => {
    mix.js("src/app2/js/index.js", "dist/app2.js")
    mix.css("src/app2/css/index.css", "dist/app2.css")
    mix.options({
        processCssUrls: false,
    })
})
```

## Manifest & Versioning

Manifest generation & versioning will be completely re-tooled to run entirely through webpack. This should eliminate any hash differences between generated chunks in webpack and when versioned separately by Mix. This also means that we will automatically merge the manifest across multiple builds when shared. However each build can have its own manifest by setting a different name via `mix.options`.

## Handling Mix Extensions / Components

This is still very much up in the air but the thought is right now that we'll instantiate an extension/component once per build group so they can keep their state local to a single group.


meshakeeb:
How soon we be seeing this in production builds?

thecrypticace:
I'm not certain at the moment. I work on Mix in my spare time. There's still plenty I need to figure out here.

meshakeeb:
I never worked on laravel mix let me know how to help you to speedup this PR I need it in one of my project

thecrypticace:
@meshakeeb heh yeah there's still quite a bit to do here. I don't have a definitive todo list but I don't expect this feature to be merged for a while. There's a lot of testing and tweaks that have to be done. It's likely this will go through a pretty heavy beta cycle as well.

I don't have a definitive todo list but it is something I need to think more about and write down

PR #3136, opened 2021-10-25
Overwrite mini-css plugin with new filename-placeholders

vilbergs:
This is unlikely to make the cut, but I wanted to at least showcase my way of solving the issues found in https://github.com/laravel-mix/laravel-mix/issues/2589, as inspiration for a more thought out approach from maintainers that know the codebase.

To summarize, the only way I could find to modify the output of the extraction was to completely disable the manually added chunks, and then (aggressively) overwrite the `MiniCssExtractPlugin`. I'm sure there is a better way, or at least more graceful one.

### DIFF
--- a/test/features/vue.js
+++ b/test/features/vue.js
@@ -19,9 +19,6 @@ test('async components', async t => {
 
     t.true(File.exists(`test/fixtures/app/dist/css/app.css`));
     t.true(File.exists(`test/fixtures/app/dist/css/async-component.css`));
-
-    // t.true(File.exists(`test/fixtures/app/dist/js/app.js`));
-    // t.true(File.exists(`test/fixtures/app/dist/css/app.css`));
 });
 
 test('it adds the Vue 2 resolve alias', async t => {


thecrypticace:
Sorry I've been a bit busy but I'm planning on taking a look at this today!

thecrypticace:
So I took a look last night and I think doing this properly may require a webpack plugin. This is something I'm going to look further into to see.

vilbergs:
> So I took a look last night and I think doing this properly may require a webpack plugin. This is something I'm going to look further into to see.

Awesome. Thanks for taking the time to look into this. We settled on using the current filepath option as the output file turned out to be reasonable in size. We did however Hijack MiniCSSExtractPlugin in a similar way to this PR so that we could ignore order warnings.

Of course it would be great to tweak the options here a little so that we could have a more sane config. But at least we've solved the issue on our end for now.

vilbergs:
Closing for now as we've found a suitable workaround. This may not be worth spending more time on.

PR #3153, opened 2021-11-07
Support ESM mix config

thecrypticace:
# This is planned for v6.1

This adds support for a few big features:
1. An ESM mix config — you can now use import/export with `type: "module"` or the extension `.mjs`.
2. A default exported function. In the absence of TLA support you can use and export a default function which can optionally be asynchronous that defines the config. It gets passed the `mix` variable/API that is exported from the package normally.
3. Along with item 2, this has enabled a `defineConfig` helper. This kind of thing has started to become a bit more common and it aids typing as well.

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -77,6 +77,10 @@ async function executeScript(cmd, opts, args = []) {
     ].join(' ');
 
     const scriptEnv = {
+        // Allow dynamic ESM imports
+        // This is an unfortunate workaround but it's the simplest way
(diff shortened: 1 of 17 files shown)


thecrypticace:
Things I didn't know about JS. Functions have an empty object prototype unless it's an arrow fn. Which has none…

🤦

thecrypticace:
Published v6.0.39 which should fix it.

JeffreyWay:
Haha, yikes, I didn't know that either.

PR #3157, opened 2021-11-09
Img conflict name

### LINKED ISSUE #3156: Bundling images with the same file name but with different paths causes an error
-   Laravel Mix Version: 6.0.0
-   Node Version: 14.17.0
-   NPM Version: 6.14.13
-   OS: MacOs Catalina

### Description:
When in the project we have multiple images that have the same filename, importing them into the project will cause an issue when bundling the project. Webpack is unable to overwrite the previously copied file to the public/images (or the dist folder) because it already exists from the first import.

aarbi:
fixes #3156

### DIFF
--- a/src/builder/webpack-rules.js
+++ b/src/builder/webpack-rules.js
@@ -32,7 +32,7 @@ module.exports = function (mix) {
                             if (!/node_modules|bower_components/.test(path)) {
                                 return (
                                     mix.config.fileLoaderDirs.images +
-                                    '/[name].[ext]?[hash]'
+                                    '/[name].[hash:8].[ext]?[hash:8]'
                                 );
                             }
 
--- a/src/components/Options.js
+++ b/src/components/Options.js
@@ -20,16 +20,14 @@ class Options {
         if ('extractVueStyles' in options) {
             messages.push({
                 type: 'warn',
-                text:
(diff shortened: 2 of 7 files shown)


aarbi:
the hashes are different from what I got in my local machine.
I haven't found a method in the File class that supports fuzzy search.

I will start by getting the hashes of files generated on the Github actions environement and stick them in the tests.
if this is not good enough, I will add a helped function that uses fs-extra to do something like this:

```
/**
 * Finds the files in a folder that have a given file name and extension.
 * This helps when looking for files generated by webpack with hashes
 *
 * @param {string} path
 * @param {string} fileName
 * @param {string} extension
 * @returns
 */
export async function filesFoundInFolder(path, fileName, extension) {
	const filesInFolder = await fs.readdir(path);
	const filteredFiles = filesInFolder.filter(
		f => f.startsWith(fileName) && f.endsWith(extension)
	);

	return filteredFiles;
}
```

aarbi:
So webpack hashes aren't generated based on the absolute paths, but from the where the root dir of the project. Which means that the hashes are consistent if the test are run in local and github actions. I set the exact has I had in local and it works now in Github. It didn't work earlier to me because I used 6 chars instead of 8

iamarpitpatidar (approved):

PR #3162, opened 2021-11-14
Test refactoring

thecrypticace:
Still a big WIP. Want to cleanup some of the assertions so they read better as well as prep for parallel testing support.


thecrypticace:
So far I've reduced the test time by quite a bit. Nearly ~2m on Linux and ~3.5m on Windows.

Next step will be trying to implement parallel testing to see if that makes any difference. It theoretically should but not sure how much of one it can/will make.

thecrypticace:
Obviously still much to do here but making progress. There are some race conditions which will need to be addressed (hence the failures) but testing is now down to around ~4m (and that win test seems to be going a good bit slower than the others for whatever reason).

PR #3208, opened 2022-02-01
Fix pattern for matching chuckHashes

tomosterlund:
Updating from version 6.0.19 to 6.0.41 breaks our entire application. The Problem:

Somewhere along the way between minor versions 19 and 41, through the Webpack dependency in Laravel mix I assume, the chunkHashes for JS-files in our project were shortened to 16 characters instead of 20 like before. For `app.js` and `manifest.js` however, the chunkHashes remain 20 characters long. It appears to me that Laravel Mix is able to cope with this, as long as I let the manifest handle all the mapping of what modules to fetch. However, if I want to prefetch JavaScript-files, or just declare them in my template (which for us drastically reduces load times), such as:

````
<link rel="prefetch" href="{{ mix('js/direction-maps.js') }}">
<script async src="{{ mix('js/detail-product-list.js') }}"></script>
````

I get the following error:

```
Exception - Unable to locate Mix file: /js/**js-file-name**.js. (View: /Users/tsd/PhpstormProjects/project-name/resources/views/components/head/prefetching.blade.php)
Unable to locate Mix file: /js/**js-file-name**.js.
```

Since the chunkHashes are sometimes 16 characters long, and sometimes 20 like before, this would be my suggested fix.

### DIFF
--- a/src/Manifest.js
+++ b/src/Manifest.js
@@ -42,7 +42,7 @@ class Manifest {
     add(filePath) {
         filePath = this.normalizePath(filePath);
 
-        let original = filePath.replace(/\?id=\w{20}/, '');
+        let original = filePath.replace(/\?id=(\w{20}|\w{16})/, '');
 
         this.manifest[original] = filePath;
 
--- a/src/Manifest.js
+++ b/src/Manifest.js
@@ -42,7 +42,7 @@ class Manifest {
     add(filePath) {
         filePath = this.normalizePath(filePath);
 
-        let original = filePath.replace(/\?id=(\w{20}|\w{16})/, '');
+        let original = filePath.replace(/\?id=\w+/, '');
 
         this.manifest[original] = filePath;
 

### CODE REVIEW

File: src/Manifest.js [hunk 1]
```diff
@@ -42,7 +42,7 @@ class Manifest {
     add(filePath) {
         filePath = this.normalizePath(filePath);
 
-        let original = filePath.replace(/\?id=\w{20}/, '');
+        let original = filePath.replace(/\?id=(\w{20}|\w{16})/, '');
```
thecrypticace, line 45:
```suggestion
        let original = filePath.replace(/\?id=\w+/, '');
```

thecrypticace:
Thanks! I *think* this is because webpack changed hash algorithms internally. The hashing in Mix is currently inconsistent and really should be handled by webpack all the way through. That will likely have to wait for a v7 release though as it could break quite a few things.

Running tests now…

tomosterlund:
Thank you. I wanted to extend a little on the tests in the extraction-Suite, but they seemed to be failing (without my changes too).

thecrypticace:
I'll tag a release soon (probably Friday). There's a few things I need to look at in issues and PRs first.

thecrypticace:
Haven't forgotten about this. Just got some life things in the way. I'll cut a release this week. Sorry for the delay!

tomosterlund:
Thanks a lot!

PR #3209, opened 2022-02-02
Fixes for new tolerance for chunkHashes

tomosterlund:
Related PR: https://github.com/laravel-mix/laravel-mix/pull/3208

In the PR yesterday, the tolerance for chunkHashes was expanded from matching `/\w{20}/`, to matching `/\w+/`. This calls for further adjustment of the version()-function, in case someone would now set the chunkHashes to be even longer, and it needed a slight adjustment of a test case.

Sorry for the commit-jumble. I forgot to sync the upstream before pushing to my fork. If this is a problem I can make a new fork and just add 1 commit instead.


thecrypticace:
No worries. I'm gonna squash merge the PR anyway :D

PR #3250, opened 2022-04-25
Use webpack asset modules

### LINKED ISSUE #2914: Use Asset Modules over file-loader
According to the webpack 5 documentation the `file-loader` should be replaced by the [asset module](https://webpack.js.org/guides/asset-modules/).
Laravel mix still uses the `file-loader` instead. Would it be a good idea to replace it then?

### LINKED ISSUE #3249: Support for Node 18
-   Laravel Mix Version: 6.0.43 (`npm list --depth=0`)
-   Node Version (`node -v`): 18.0.0
-   NPM Version (`npm -v`): 8.7.0
-   OS: irrelevant

### Description:
Building fails on Node v18, due to the upgrade to OpenSSL v3 (since Node 17). In #3138 and #3140 some fixes were done to ensure compatibility, but they do very specific matching on the version number.

### Steps To Reproduce:
Running test suite on Node 18 should probably be enough.

### Possible solutions:
- Simply adding an additional `process.version.startsWith('v18.')` wherever applicable (note that this means things will still break again on next Node releases)
- Doing some kind of semver-range matches on all places where Node 17 is now explicitly matched

jnoordsij:
Replace deprecated `file-loader` with Webpack asset modules, as described in https://webpack.js.org/guides/asset-modules/.

Fixes #2914 and fixes #3249 for my setup.

### DIFF
--- a/src/builder/webpack-default.js
+++ b/src/builder/webpack-default.js
@@ -20,6 +20,7 @@ module.exports = function (mix) {
         entry: {},
 
         output: {
+            assetModuleFilename: '[name][ext]?[hash]',
             chunkFilename: '[name].[hash:5].js'
         },
 
--- a/src/builder/webpack-rules.js
+++ b/src/builder/webpack-rules.js
@@ -6,6 +6,37 @@ module.exports = function (mix) {
     // TODO: Remove in Mix 7 -- Here for backwards compat if a plugin requires this file
     mix = mix || global.Mix;
 
+    function buildRule({ test, name = null, loaders = [] }) {
+        if (mix.config.assetModules) {
+            return {
+                test,
+                type: 'asset/resource',
+                generator: {
+                    filename: (pathData) => name(pathData, { dirs: mix.config.assetDirs }),
+                    publicPath: mix.config.resourceRoot
+                },
+                use: loaders,
+            }
+        }
+
+        return {
+            test,
+            use: [
+                {
+                    loader: mix.resolve('file-loader'),
+                    options: {
+                        // we're somewhat mimic-ing the asset module API here to simply name resolution further down
+                        name: (path) =>
+                            name({ filename: path }, { dirs: mix.config.fileLoaderDirs || mix.config.assetDirs })
+                                .replace('[ext]', '.[ext]'),
+                        publicPath: mix.config.resourceRoot
+                    }
+                },
+                ...loaders,
+            ]
+        }
+    }
+
     /** @type {import("webpack").RuleSetRule[]} */
     let rules = [];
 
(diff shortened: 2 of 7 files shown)


thecrypticace:
Because people are fairly reliant on `file-loader` we'll want a way to _opt-in_ to asset modules. I suggest adding a config option for this (so you can enable it via `mix.options`). You'll be able to pull from the config there using `mix.config`.

Maybe something like `mix.options({ assetModules: true })` though I'm very open to other suggestions for naming.

thecrypticace:
I do want to _require_ them in the next major version. It should've been caught when we upgraded Mix to v6 but sadly I missed it. 🙃

jnoordsij:
Thanks for your feedback! I've added a config option now that toggles the new behavior. Also I adjusted tests to use this new toggle by default and copied the two main tests which test these rules to test both old and new version. Finally force-pushed to remove the commit that deleted the `file-loader` dependency, as it is still used.

Let me know if you have any more suggestions and/or improvements!

thecrypticace:
Used this as an opportunity to do a tiny refactor of the rules code. I've always felt it was far too messy. I _think_ it's better but it doesn't feel 100% where I want it to be. But better is better in any case. Thanks again for doing this! ❤️ 

I think there's a few other things I'll want to take care of before tagging a patch release but I'm aiming to cut a release on either Friday or Saturday.

jnoordsij (commented):
Great, thanks for merging. Looking forward to the release! Noticed two possibly incorrect references, might wanna check/fix those first.

jnoordsij:
> Used this as an opportunity to do a tiny refactor of the rules code. I've always felt it was far too messy. I _think_ it's better but it doesn't feel 100% where I want it to be. But better is better in any case. Thanks again for doing this! ❤️
> 
> I think there's a few other things I'll want to take care of before tagging a patch release but I'm aiming to cut a release on either Friday or Saturday.

Any update on the next patch release? 😇

thecrypticace:
Heh yeah sorry! Life got in the way. Soon™ — I want to see if I can fix the `[object Module]` in Vue 2 with these first.

jnoordsij:
> Heh yeah sorry! Life got in the way. Soon™ — I want to see if I can fix the `[object Module]` in Vue 2 with these first.

Great! Would be nice to have before June 1st, when Node 17 reaches EOL, so I can upgrade my builds to Node 18 by then.

thecrypticace:
Ah yesss it does fix it! 💯 

Just more evidence that I really shouldn't have missed this the first time around when I upgraded Mix to webpack 5. lol

PR #3262, opened 2022-05-21
Added online version of autoprefixer

rahulkumarsingh73690:
Added online version of Autoprefixer. So users can instantly use it without installing it.


thecrypticace:
Appreciate the PR but not sure this is really relevant for the Mix docuementation. Mix is intended to be used as a build tool and for people to install autoprefixer alongside it — in which case it'll get run automatically.

Thank you though!

PR #3263, opened 2022-05-23
Fixes Autofixer warning about deprecated color-adjust replaced by print-color-adjust

alangiacomin:
Fixes Autofixer warning about deprecated color-adjust replaced by print-color-adjust
https://github.com/postcss/autoprefixer/issues/1454


ankurk91:
i don't think it will remove the warning

alangiacomin:
> i don't think it will remove the warning

You're right, doing more deep investigation I found another package with older autofixer version.
I'll close che PR.
Thanks.

PR #3275, opened 2022-06-08
Fix Webpack 'vue-loader' Module not found error

BrandonSurowiec:
Every time that I upgrade one of my older projects to Laravel Mix 6 I have run into this MODULE_NOT_FOUND bug due to version 15.9.7 of `vue-loader` being installed. This bumps the default version of `vue-loader` for Vue 2 to 15.9.8.

Version 15.9.7 has a bug where Webpack cannot find the module:
https://github.com/vuejs/vue-loader/issues/1859

Version 15.9.8 fixes that:
https://github.com/vuejs/vue-loader/releases?q=v15.9.8&expanded=true


thecrypticace:
Thanks!

PR #3327, opened 2022-11-28
Update tests to run on active Node versions

jnoordsij:
Node v12 and v17 have reached EOL; Node v19 has been released.

Note: it might be desirable to also update the required Node.js version in `package.json` in a future version.

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -20,7 +20,7 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                node-version: [12.x, 16.x, 17.x, 18.x]
+                node-version: [14.x, 16.x, 18.x, 19.x]
                 os: [ubuntu-latest, windows-latest]
 
         steps:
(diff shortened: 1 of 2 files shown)


thecrypticace:
Hey, sorry! Life and work generally has had me focused on other stuff. This looks good. I would say _if_ you can you should update to Laravel Vite because the DX is imo significantly better (and **much faster**). But definitely recognize the need for Mix still exists. I need to dedicate some time to Mix again over the weekends.

jnoordsij:
Thanks for your response! No worries here, I totally understand this project is not a main priority anymore.

I do definitely intend to migrate at some point, but still looking to find some spare time for it and also waiting for some Vite plugins (like https://github.com/ckeditor/vite-plugin-ckeditor5/) to become slightly more stable. It's great to still have this maintained as a stable backup in the meantime ;)

xdavidwu:
Can we get a new release? We need the changes to `requiresLegacyOpenSSLProvider()` of this patch for migrating old projects to newer nodejs. (We can add `--openssl-legacy-provider` to `NODE_OPTIONS` by ourselves, but having this workaround implemented on the tool itself is still better.)

PR #3328, opened 2022-11-28
Update to webpack-cli v5

jnoordsij:
Webpack CLI v5 has been released, see https://github.com/webpack/webpack-cli/releases/tag/webpack-cli%405.0.0.

The main breaking change is dropping support for Webpack v4, but as this package is already based on Webpack v5, this should not cause any problems.


jnoordsij:
I don't expect this to ever land anymore.

